Issue #25, opened 2022-07-14
--cached option for 'clean' command


yegor256:
The option `--cached` must delete `~/.eo` directory too

yegor256:
@maxonfjvipon maybe you can help with this one too?

maxonfjvipon:
@yegor256 I'll try it out

maxonfjvipon:
@yegor256 I think it can be closed

yegor256:
@maxonfjvipon thanks for your help!

PR #27, opened 2022-07-16
#25 added --cached option for clean command

maxonfjvipon:


maxonfjvipon:
@yegor256 please, check this out

maxonfjvipon:
@yegor256 please, check this out, should work

maxonfjvipon:
@yegor256 please check this out, should work

yegor256:
@maxonfjvipon the problem here is that `--cached` will be available for _all_ commands, but we want it to be available _only_ for `clean`. In other words, `--cached` should not print in the default list of options. It should print only when I say:

```
$ eoc clean --help
```

maxonfjvipon:
@yegor256 got you

maxonfjvipon:
@yegor256 I think that's it

test/commands/test_clean.js
```diff
@@ -25,18 +25,22 @@
...
+    fs.rmSync(eo, {recursive: true, force: true});
     fs.mkdirSync(path.resolve(home, 'src'), {recursive: true});
+    fs.mkdirSync(eo, {recursive: true});
     fs.writeFileSync(path.resolve(home, 'src/simple.eo'), '[] > simple\n');
     const stdout = runSync([
-      'clean', '-s', path.resolve(home, 'src'), '-t', path.resolve(home, 'target'),
+      'clean', '-s', path.resolve(home, 'src'), '-t', path.resolve(home, 'target'), '--cached',
     ]);
-    assert(!fs.existsSync(path.resolve(home, 'target')), stdout);
+    assert(!fs.existsSync(path.resolve(home, 'target')) && !fs.existsSync(eo), stdout);
```
yegor256, line 43: @maxonfjvipon if this assertion fails, it won't be possible to understand what exactly was the problem: the `target` is absent or `~/.eo`? It's better to have individual `assert` statements per each particular error you are checking.

src/eoc.js
```diff
@@ -69,10 +69,12 @@ program.command('foreign')
...
   });
 
-program.command('clean')
+program
+  .command('clean')
+  .option('--cached', 'delete ~/.eo directory')
   .description('delete all temporary files')
   .action((str, opts) => {
-    clean(program.opts());
+    clean({...str, ...program.opts()});
```
yegor256, line 77: @maxonfjvipon I have a strong feeling that it's wrong. Try to run this and you will the problem:

```
$ eoc clean --cached --foo
```

yegor256:
@maxonfjvipon see above

src/eoc.js
maxonfjvipon: with `--foo` we get `error: unknown option '--foo'`, so it works absolutely correctly, but not quite obviously, will fix it.

test/commands/test_clean.js
maxonfjvipon: fixed

src/eoc.js
maxonfjvipon: fixed

maxonfjvipon:
@yegor256 please check one more time

yegor256:
@rultor merge

yegor256:
@maxonfjvipon did you try `grunt eslint`?

maxonfjvipon:
@yegor256 please try merge one more time

yegor256:
@rultor merge

yegor256:
@rultor merge

Issue #88, opened 2022-10-28
Increase of stack size


ivan-egorov42:
We need to increase stack size. According to objectionary/eo/issues/1278

ivan-egorov42:
@Graur pls, assign me.

yegor256:
@rultor release, tag is `0.9.0`

PR #89, opened 2022-10-28
Added option for changing stack size

ivan-egorov42:
According to #88.

### DIFF
--- a/src/commands/dataize.js
+++ b/src/commands/dataize.js
@@ -36,6 +36,7 @@ module.exports = function(obj, args, opts) {
     `java`,
     [
       '-Dfile.encoding=UTF-8',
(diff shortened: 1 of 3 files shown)


ivan-egorov42:
@Graur  @mximp pls, have a look. I changed `test_dataize` in order to check the functionality of stack size option of a compiler. I'm not sure if we need separate test for it, but if we need, I will add todo for it. 

I added a long list to the test for checking StackOveflow. Now we don't have StackOveflow in test due to stack size option which increases size of the stack.

Graur (approved):
@yegor256 please, have a look

yegor256:
@rultor merge

yegor256:
@ivan-egorov42 thanks!

Issue #102, opened 2022-12-10
EO version noise


kerelape:
"EO version is 0.28.11" is printed after each step when run `eoc dataize main`.

```
EO version is 0.28.11
EO objects registered in /home/kerelape/Documents/cir/.eoc/eo-foreign.json
EO version is 0.28.11
EO program assembled in /home/kerelape/Documents/cir/.eoc
EO version is 0.28.11
Java sources generated in /home/kerelape/Documents/cir/.eoc/generated-sources
EO version is 0.28.11
Java .class files compiled into /home/kerelape/Documents/cir/.eoc
EO version is 0.28.11
Executable JAR created at /home/kerelape/Documents/cir/.eoc/eoc.jar
```

yegor256:
@volodya-lombrozo really annoying bug, we should somehow introduce global caching

PR #105, opened 2022-12-14
feat(#102): return cached eo plugin version

volodya-lombrozo:
Use cache for version instead of calculating it each time.

Closes: #102


volodya-lombrozo:
@yegor256 have a look

yegor256:
@rultor merge

Issue #368, opened 2024-11-04
we don't show timing in the main log output


yegor256:
Currently, I'm getting this:

```
$ eoc compile
EO objects registered in .eoc/eo-foreign.json
EO program assembled in .eoc
EO program verified in .eoc
Java sources generated in .eoc/generated-sources
Java .class files compiled into .eoc
```

Would be much better to see this:

```
EO objects registered in .eoc/eo-foreign.json in 34ms
EO program assembled in .eoc in 1s
EO program verified in .eoc in 340ms
Java sources generated in .eoc/generated-sources in 5s
Java .class files compiled into .eoc in 2s
```

Let's add the timings at the end of every logging line.

ReBy298:
I would like to contribute to this issue :D

yegor256:
@ReBy298 please, go head!

0pdd:
@yegor256 2 puzzles [#434](https://github.com/objectionary/eoc/issues/434), [#435](https://github.com/objectionary/eoc/issues/435) are still not solved.

0pdd:
@yegor256 all 2 puzzles are solved here: [#434](https://github.com/objectionary/eoc/issues/434), [#435](https://github.com/objectionary/eoc/issues/435).

PR #372, opened 2024-11-07
Feat: Add timing to the main log output (ref #368)

ReBy298:
Feature #368 
Add Timing to the main log output 

Description

1. **Start Time Capture**: I introduced a startTime variable to record the current time at the beginning of the function with Date.now().

2. **Execution Duration Calculation**: After the command completes, I added a duration variable to calculate the time taken by subtracting startTime from the current time (Date.now()), which gives the duration in milliseconds.

3. **Enhanced Logging**: I updated the console.info log message to include the duration in milliseconds (%dms). Now, the log shows not only the completion message ("EO program assembled") but also the time it took to execute.

Modified lines: 
1. console.info('EO program assembled in %s in %dms', rel(target), duration); at assemble.js
2. console.info('Java .class files compiled into %s in %dms', rel(target), duration); at compile.js
3. console.info('Java sources generated in %s in %dms', rel(sources), duration); at transpile.js
4. console.info('EO objects registered in %s in %dms', rel(foreign), duration); at register.js
5. console.info('EO program verified in %s in %dms', rel(path.resolve(opts.target)), duration); at verify.js

---

## PR-Codex overview
This PR adds timing measurements to various command functions in the codebase, allowing for better performance tracking by logging the duration of each operation.

### Detailed summary
- Added `startTime` variable to capture the start time for each command function.
- Calculated `duration` by subtracting `startTime` from the current time.
- Updated console log messages to include the duration in milliseconds for:
  - `verify`
  - `assemble`
  - `register`
  - `compile`
  - `transpile`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

### DIFF
--- a/src/commands/assemble.js
+++ b/src/commands/assemble.js
@@ -33,8 +33,10 @@ const {mvnw, flags} = require('../mvnw');
  */
 module.exports = function(opts) {
   const target = path.resolve(opts.target);
+  const startTime = Date.now();
   return mvnw(['eo:assemble'].concat(flags(opts)), opts.target, opts.batch).then((r) => {
-    console.info('EO program assembled in %s', rel(target));
+    const duration = Date.now() - startTime;
+    console.info('EO program assembled in %s in %dms', rel(target), duration);
     return r;
   });
 };
--- a/src/commands/java/compile.js
+++ b/src/commands/java/compile.js
@@ -33,8 +33,10 @@ const path = require('path');
  */
 module.exports = function(opts) {
   const target = path.resolve(opts.target);
+  const startTime = Date.now();
   return mvnw(['compiler:compile'].concat(flags(opts)), opts.target, opts.batch).then((r) => {
-    console.info('Java .class files compiled into %s', rel(target));
+    const duration = Date.now() - startTime;
+    console.info('Java .class files compiled into %s in %dms', rel(target), duration);
     return r;
   });
 };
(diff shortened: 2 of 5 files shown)


yegor256:
@ReBy298 looks very good, thanks! However, don't you think that it's an obvious code duplication? There are more places in the application that need similar time reporting. If you do the same changes in them too, you will see how many duplicate lines of code we'll get. Maybe it would be better if we introduce some sort of a wrapper (this is pseudo-code):

```
  const target = path.resolve(opts.target);
  const startTime = Date.now();
  return mvnw(['eo:assemble'].concat(flags(opts)), opts.target, opts.batch).then((r) => {
    elapsed(() => {
      const duration = Date.now() - startTime;
      return `EO program assembled in ${rel(target)}`;
    }
  });
};
```

ReBy298:
> @ReBy298 looks very good, thanks! However, don't you think that it's an obvious code duplication? There are more places in the application that need similar time reporting. If you do the same changes in them too, you will see how many duplicate lines of code we'll get. Maybe it would be better if we introduce some sort of a wrapper (this is pseudo-code):
> 
> ```
> module.exports = function(opts) {
>   const target = path.resolve(opts.target);
>   elapsed((e) => {
>     return mvnw(['eo:assemble'].concat(flags(opts)), opts.target, opts.batch).then((r) => {
>       e.print(`EO program assembled in ${rel(target)}`);
>       return r;
>     };
>   });
> };
> ```
> 
> Here, the `elapsed` function is the one we create and re-use in all places, where we need to print the elapsed time.

Okey let me fix it

PR #395, opened 2024-11-26
Autodocs added

SenjeyB:
Added an automatic documantation generator.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,9 @@
         "relative": "3.0.2",
         "semver": "7.6.3",
         "sync-request": "6.1.0",
-        "xmlhttprequest": "1.8.0"
+        "xmlhttprequest": "1.8.0",
+        "marked": "15.0.2",
+        "xml2js": "0.6.2"
       },
       "bin": {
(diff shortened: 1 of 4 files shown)


yegor256:
@SenjeyB thanks for your contribution, but it's better to maintain the structure of JS files that already exists in this repo. See, how commands are dispatched from the `eoc.js` file. Your command should be `docs`. I should be able to run `eoc docs` to generate documentation. Maybe you can create the first PR with a blank implementation of the command -- it will not generate any documentation, but will work and tests will pass. Then,  you continue in the second PR, and so on.

yegor256:
@maxonfjvipon please take a look

yegor256:
@SenjeyB many thanks for your contribution! I see the following issues: 

1) keeping CSS inside JS is not a good idea, better keep it outside as a `.css` file (or even LESS or SASS)

2) rendering HTML as a text is not a good idea, better use XSLT

3) making such a big pull request may lead to much longer review cycles. Better send a shorter change, which will be easier to merge. Later, send improvements and refinements. Check this out: https://www.yegor256.com/2010/03/04/pdd.html

Issue #411, opened 2024-12-07
Problem with grunt


RealVlad322:
Hello,
This might only apply to me, but I needed to run `npm install -g grunt-cli` before run `grunt` would work.
I believe this should be mentioned in the README.

yegor256:
@RealVlad322 this is exactly how it works in CI: https://github.com/objectionary/eoc/blob/master/.github/workflows/grunt.yml#L51-L53 maybe you know how to fix this and get rid of this extra step?

RealVlad322:
what about this solution, is it right for you?
https://github.com/lastboy/package-script

yegor256:
@RealVlad322 looks like a hack, not a solution

RealVlad322:
Ok, I'll try to find another solution

Dirakon:
You can also use `npx grunt` instead of installing something globally

yegor256:
@Dirakon fixed already

PR #412, opened 2024-12-07
fix(#411)

RealVlad322:
add installer.js and rewrtite default npm i


yegor256:
@RealVlad322 thanks for your contribution! However, the build fails. Also, are you sure this is the right solution for this problem? Seems like a rather dirty hack.

yegor256:

PR #456, opened 2025-01-26
create inspect command

ErnestMatskevich:
Registered inspect comman in eoc.js
Made inspect.js command
Made tests

### DIFF
--- a/src/commands/inspect.js
+++ b/src/commands/inspect.js
@@ -1,0 +1,67 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2022-2024 Objectionary.com
+ *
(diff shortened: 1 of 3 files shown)


0crat:
@ErnestMatskevich Hey there! 👋 Just a heads up about your Git branch naming - "`inspect`" doesn't quite hit the mark according to our team's convention. [This article](https://www.yegor256.com/2014/04/15/github-guidelines.html) explains why it's not ideal. Our policy suggests using the ticket number instead, so "`455`" would've been perfect here. I know it's a small thing, but it does come with a -12 point penalty. No worries though, it's an easy fix for next time! Let's keep our branches tidy and aligned with the ticket numbers. 😊

Issue #460, opened 2025-02-14
add `generate_comments` command


Dirakon:
`eoc generate_comments` would look for all comment placeholders in the code (like `# <COMMENT-TO-BE-ADDED>`) and use LLM to replace them with proper comments.

Proposed cli interface would look like this:
```bash
eoc generate_comments\
  --provider=ollama\
  --source=app.eo\
  --comment_placeholder="<COMMENT-TO-BE-ADDED>"\
  --output=out.json\
  --prompt_template=prompt_template.txt
```

Additionally, some provider-specific command line arguments would also be present.

The output file would contain the corresponding comment for each placeholder.

Dirakon:
Readme still needs to be appended

PR #461, opened 2025-02-14
feat(#460): Add `generate_comments` command with placeholder and openai providers

Dirakon:
- Added langchain package for support of different LLM providers
- Added support for placeholder provider with CLI interface proposed in the issue.
- Added some tests for placeholder provider.
- Added langchain-openai provider

PR #496, opened 2025-04-15
Draft: feat(#460): install langchain

Dirakon:
Part of implementation of #460 

Added langchain package for support of different LLM providers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,11 +15,12 @@
         "win32"
       ],
       "dependencies": {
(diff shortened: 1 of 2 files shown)


yegor256:
@Dirakon this change is not motivated enough. Better submit it together with the feature that depends on this package.

Dirakon:
I see. In that case, this merge request should work better - https://github.com/objectionary/eoc/pull/461

Issue #500, opened 2025-04-23
`inspect` command


yegor256:
When I start `eoc inspect` it should load all `.class` files into the VM and then give me an interactive interface, for example:

```
$ eoc inspect
Loaded 56 objects
Ready to traverse the Universe
@ Φ (5 attrs)
_
```

Current position in the tree is printed with the `@` prefix.

Then, I type in:

```
@ Φ ⟦org⟧
.org

@ Φ.org ⟦bytes⟧
.eolang

@ Φ.org ⟦number bytes go malloc ... +15⟧
.bytes

@ Φ.org.bytes ⟦data and or xor div plus minus ... +7⟧
ls
.data ↦ ∅
.and ↦ ⟦ λ, b ⟧
.or ↦ ⟦ λ, b ⟧
...

@ Φ.org.bytes ⟦data and or xor div plus minus ... +7⟧
cp Φ.bar
Copied @ Φ.org.bytes to @ Φ.bar

@ Φ.bar ⟦data and or xor div plus minus ... +7⟧
ls
.data ↦ ∅
.and ↦ ⟦ λ b ⟧
.or ↦ ⟦ λ b ⟧

@ Φ.bar ⟦data and or xor div plus minus ... +7⟧
.data

@ Φ.bar.data ∅
form

@ Φ.bar.data ⟦⟧
put 04-05-96-92-2F-E3

@ Φ.bar.data ⟦Δ⟧
run
04-05-96-92-2F-E3
```

There are four possible commands:

* `..` - jump to the upper object in the tree
* `add foo` (or `+foo`) - add new attribute `foo` to the current object
* `rm foo` (or `-foo`) - delete `foo` attribute from the current object
* `go foo` (or `.foo`) - jump to the object attached to the `foo` attribute
* `to Φ.bar` - attach `Φ.bar` to the current object
* `form` - create an empty formation and attach to the current object
* `put 75-E4-3F` - attach data to the `Δ` asset of the current object
* `cp Φ.bar` - make a copy of the current object and attach it to `Φ.bar`
* `dd foo Φ.bar` - dispatch `foo` and attach the result to `Φ.bar`
* `ls` - print all attributes of the current object
* `dataize` (or `run`) - dataize current objects and print result

yegor256:
@maxonfjvipon @volodya-lombrozo what do you think?

volodya-lombrozo:
@yegor256 

1. Do you want to add the `inspect` command? Or it exists already? 
[...]

PR #528, opened 2025-05-22
500, adding entry point to inspect command

ErnestMatskevich:
Renamed branch to #500 for tracking issue
Added test for Inspect.java file


ErnestMatskevich:
I added simple test for `Inspect.java` file, but when run qulice I faced with tha same warnings as in previous PR:

```
[WARNING] Unused declared dependencies found:
org.eolang:eo-runtime:jar:0.56.2:compile
org.junit.jupiter:junit-jupiter-api:jar:5.12.2:compile
org.junit.jupiter:junit-jupiter-engine:jar:5.12.2:compile
org.takes:takes:jar:1.24.6:compile
```

Also I would like to clarify how to build `inspect.jar` with Maven. Since currently there is single `pom.xml` file, Maven makes `eoc.jar` file as it writen in its <artifactId>. For complete my task I need to create my own file `inspect.jar`. I would like to use `<execution>` in `pom.xml` to separate creating `eoc.jar` and my `inspect.jar`. Is it correct idea? I think that after separate creating `inspect.jar` there will be no more warning about unused dependencies, because qulice will be placed in `<execution>` of inspect part and will not see junit and eo-runtime dependecies. @yegor256 What do you think about use _<execution>_ in `pom.xml`?

There is 1 falling check in [eslint workflow](https://github.com/objectionary/eoc/actions/runs/15198624288/job/42748385329?pr=528) Is it on my side? If it is how can I fix that?

yegor256:
@ErnestMatskevich I fixed eslint, please merge master into your branch

yegor256:
@volodya-lombrozo can you please check this design?

volodya-lombrozo (changes requested):
@ErnestMatskevich 

Thank you for the contribution! I have a few important notes related to the PR itself. Highly recommend reading ["Helping others review your changes"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/helping-others-review-your-changes). Pay attention to  the [Write small pull requests](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/helping-others-review-your-changes#write-small-pull-requests) and the [Provide context and guidance](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/helping-others-review-your-changes#provide-context-and-guidance) sections. Could you also please make the title of this PR a bit more meaningful, rather than just "500", please?

volodya-lombrozo (changes requested):
@yegor256 

### `inspect` -> `debug`

As I understand it, you want to write a debugger. Why not call it `eoc debug` then? Why should we hide this information under the misleading `inspect` command?

## Separate Project

Why do we need to implement the entire debugger in the `eoc` repository? Maybe it's better to move it to a separate repo, as [other projects](https://github.com/go-delve/delve) do?
Here, we can just leave a thin `eoc debug` command just to invoke this debugger.
Moreover, `eoc` is written in JS.  Adding `Java` would create a some sort of mess. 

What do you think?

yegor256:
@volodya-lombrozo how would you call a tool that allows you go into a working Java virtual machine, find any object and see its details? Would it be a debugger?

Indeed, this tool deserves a separate project/repository, but maybe a bit later, when its first version is ready. So far, it's an early draft.

ErnestMatskevich:
@volodya-lombrozo Thank you for your review feedback. I've made the following improvements:

1. Updated PR title to clearly state the functionality.
2. Will follow up with smaller, focused PRs containing detailed descriptions

For the current changes in [pom.xml](https://github.com/objectionary/eoc/blob/5e205ea31878c21b84df4216f7ceb936213b3d59/mvnw/pom.xml) added Maven profile section to build self-contained `inspect.jar`. 

inspect command can be run with help 
```bash 
node src/eoc.js inspect 
``` 
or it is possible just build `inspect.jar` with help Maven and use `-Pinspect` for indepening package/compile.

Could you please review this profile configuration approach.

volodya-lombrozo (changes requested):

src/commands/inspect.js
```diff
@@ -0,0 +1,78 @@
...
+const { spawn } = require('child_process');
+const path = require('path');
+const os = require('os');
+const readline = require('readline');
+const fetch = require('node-fetch');
+
+module.exports = function(opts) {
+  return new Promise((resolve, reject) => {
+    const mvnDir = path.join(process.cwd(), 'mvnw');
+    console.info('Building EO inspect server with Maven...');
```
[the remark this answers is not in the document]
yegor256: @volodya-lombrozo @ErnestMatskevich indeed, it's better to build it once, when we package eoc. Then, simply run it.

ErnestMatskevich:
@volodya-lombrozo I implemented 4 out of 5 feedback items from your review. Below is the current progress:

**Completed:**

- Added asserts in `InspectSmokeTest.java` at [this commit](https://github.com/objectionary/eoc/pull/528/commits/e9382b508a7ceda29f2935d91df4f0557cfd827d)
Also changed `Inspect.java` to handle interrupt server

- Removed redundant `<resources>` in `pom.xml` and `qulice` plugin (will add in new PR) at [this commit](https://github.com/objectionary/eoc/pull/528/commits/057fee345d010edcfa718a347a66855e57163f0d)

- Simplified variable names in `test_inspect.js` at [this commit](https://github.com/objectionary/eoc/pull/528/commits/314b44d33089b34424d914ad8672567d85b10220)
Fixed object-shorthand for ESLint at [last commit](https://github.com/objectionary/eoc/pull/528/commits/59df1e8df75851397ec3e6894692f185df657ca2)

**In progress:**
- Change Maven configuration to build java server once

Could you please review the implemented changes?

volodya-lombrozo (changes requested):

mvnw/src/main/java/org/eolang/Inspect.java
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * HTTP inspection server.
+ * @since 0.29.0
+ */
+public final class Inspect {
+    /**
+     * Private constructor.
+     */
+    private Inspect() {
```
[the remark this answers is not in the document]
ErnestMatskevich: @volodya-lombrozo I added `private` to constructor to follow `qulice` rules when qulice plugin was part of my code. Without private constructor, qulice spawned this error: 

All methods are static.  Consider using a utility class instead. Alternatively, you could add a private constructor or make the class abstract to silence this warning. (UseUtilityClass)

I agree that it is not utility class since there is a `main()`. I deleted misunderstinding comment and saved constructor itself to follow `qulice` rule in future, when plugin will be added in project.

All changes regarding to Inspect.java file were done in [this commit](https://github.com/objectionary/eoc/pull/528/commits/969012b59f9a7b9e95203347ec819be010db9895)

ErnestMatskevich:
@yegor256 @volodya-lombrozo I changed `eoc inspect` command behaviour. Now command `mvn clean package` creates `inspect.jar` with server and `takes` dependency inside. `inspect.js` just load `inspect.jar` and does not build it every time. Done in [commit](https://github.com/objectionary/eoc/pull/528/commits/3e593de7135169dc3f88a07a6ff08e7e8843f613)

I wrote descriptor `src/assembly/inspect.xml` for `maven-assembly-plugin` which I use for build `inspect.jar` and excluded `.class` files with my code in `maven-jar-plugin` for `eoc.jar`.

I deleted Maven part from inspect.js since `inspect.jar` is building by `mvn clean package` and deleted checks for Maven in `test_inspect.js`


There are problems in workflow with build and itests:
- ### Build fails
1) [dataize at 0.56.2 version](https://github.com/objectionary/eoc/actions/runs/15377766908/job/43264609698?pr=528#step:7:222)
2) [dataize with command-line argument](https://github.com/objectionary/eoc/actions/runs/15377766908/job/43264609698?pr=528#step:7:235)
dataization for older version is passed, but this 2 steps fails. Is problem on my side? I did not edit default building except adding exclusions for files with my code to avoid include them in eoc.jar 

- ### itests fails
[Could't find object 'Φ.org.eolang.io'](https://github.com/objectionary/eoc/actions/runs/15377766911/job/43264609705?pr=528#step:6:29)

Could you please take a look for my [commit](https://github.com/objectionary/eoc/pull/528/commits/3e593de7135169dc3f88a07a6ff08e7e8843f613) and help to fix build

mvnw/src/main/java/org/eolang/Inspect.java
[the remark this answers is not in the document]
volodya-lombrozo: @ErnestMatskevich Great thanks for the explanation. Now it's clear.

mvnw/src/test/java/org/eolang/InspectTest.java
```diff
@@ -0,0 +1,66 @@
...
+        final long start = System.currentTimeMillis() + 10_000L;
+        while (System.currentTimeMillis() < start && !server.isAlive()) {
+            Thread.sleep(100);
+        }
+        assertThat(
+            "Server thread should be running after start",
+            server.isAlive(),
+            is(true)
+        );
+
```
volodya-lombrozo, line 54: @ErnestMatskevich Can we remove empty lines, please? It's a [code smell](https://www.yegor256.com/2014/11/03/empty-line-code-smell.html).

mvnw/src/test/java/org/eolang/InspectTest.java
```diff
@@ -0,0 +1,66 @@
...
+        
+        server.start();
+        
+        final long start = System.currentTimeMillis() + 10_000L;
+        while (System.currentTimeMillis() < start && !server.isAlive()) {
+            Thread.sleep(100);
+        }
+        assertThat(
+            "Server thread should be running after start",
+            server.isAlive(),
```
volodya-lombrozo, line 51: @ErnestMatskevich Here you just check the the thread is alive. It tells nothing about the server availability.
https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#isAlive--

mvnw/src/test/java/org/eolang/InspectTest.java
```diff
@@ -0,0 +1,66 @@
...
+                public void run() {
+                    try {
+                        Inspect.main();
+                    } catch (final IOException ex) {
+                        throw new IllegalStateException("IOException while starting server", ex);
+                    }
+                }
+            }
+        );
+        server.setDaemon(false);
```
volodya-lombrozo, line 41: @ErnestMatskevich Why do we need this? From https://stackoverflow.com/a/17878439/10423604:

> Any thread created by main thread, which runs main method in Java is by default non daemon because Thread inherits its daemon nature from the Thread which creates it i.e. parent Thread and since main thread is a non daemon thread, any other thread created from it will remain non-daemon until explicitly made daemon by calling setDaemon(true).

volodya-lombrozo (changes requested):
@ErnestMatskevich There are significant improvements. Thanks! Just a few small comments. 

As for filling build, I suggest creating a new issue for that.

mvnw/src/test/java/org/eolang/InspectTest.java
ErnestMatskevich: @volodya-lombrozo Yes, the purpose of this test now is just check that server is starting without any errors. The availability of server is cheking in `test_inspect.js` file.

volodya-lombrozo: @ErnestMatskevich How do you know there aren't errors? It may be that the server didn't actually start, but the thread is still alive.

ErnestMatskevich:
### Fixed all errors

1) Updated `package-lock.json` for includes `node-fetch` - [commit](https://github.com/objectionary/eoc/pull/528/commits/b6fca8b73cb3a0bd86b7e03a6c7dd4089194e755)

2) Set `src/main/java` in `<sourceDirectory>` of `<build`> section in `pom.xml` for help Maven to see Java code of Inspector and avoid errors with unrecognising Java code, when run other commands which set different directories in `<sourceDirectory>`. For takes all previous functionality of other part of `eoc` source code, in `build-helper-maven-plugin` from `pom.xml` was set original `<source>${eo.generatedDir}</source>`. - [commit](https://github.com/objectionary/eoc/pull/528/commits/ca58081f354608a1f9d8d533443e14e2e44da065)

3) The main problem was this line: `<exclude>**/package-info.class</exclude>` in `pom.xml`. 
The `**` part exclude all `package-info.class` in project, but should excludes specific` package-info.class` for Inspect part.
After refactoring line is: `<exclude>org/eolang/package-info.class</exclude>` and all checks are passed. - [commit](https://github.com/objectionary/eoc/pull/528/commits/5be56ca094c157cbf5bff36916b7e06337996992)

@yegor256 Please, take a look

yegor256:
@ErnestMatskevich we don't know for how long these files will stay with us. Once they are merged, they are with us (forever). It's better to place them correctly. The `mvnw/` is not a "Maven folder". It's "Maven wrapper". The files you commit have nothing to do with the wrapper, right?

ErnestMatskevich:
> @ErnestMatskevich we don't know for how long these files will stay with us. Once they are merged, they are with us (forever). It's better to place them correctly. The `mvnw/` is not a "Maven folder". It's "Maven wrapper". The files you commit have nothing to do with the wrapper, right?

Done in this [commit](https://github.com/objectionary/eoc/pull/528/commits/956fa4f5a052136004d01d716c39ccd1c1857462)

Yes, I do not work with wrapper directly, i just build `JAR` file for server

volodya-lombrozo (changes requested):

ErnestMatskevich:
@volodya-lombrozo Please, take a look. All tasks done, all checks passed

package.json
```diff
@@ -26,8 +26,12 @@
     "grunt-contrib-clean": "^2.0.1",
     "grunt-eslint": "^25.0.0",
     "grunt-mocha-cli": "^7.0.0",
-    "mocha": "^11.5.0",
-    "patch-package": "^8.0.0"
+    "mocha": "10.8.2",
```
Favixx, line 29: `mocha` is declared three times in `devDependencies` here (lines 29, 32, 34: `10.8.2`, `11.5.0`, `^11.5.0`). The last one silently wins, so the first two are dead — this looks like an unresolved merge conflict. Please keep a single `mocha` entry (`^11.5.0`) and drop the other two.

src/commands/inspect.js
```diff
@@ -0,0 +1,57 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2025 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+const { spawn } = require('child_process');
+const path = require('path');
+const os = require('os');
+const readline = require('readline');
+const fetch = require('node-fetch');
```
Favixx, line 9: `node-fetch` is `require`d here but it is **not** listed in `dependencies` in `package.json` — it only resolves today because it happens to be hoisted as a transitive dep (`^2.6.7` in the lockfile). This will break the moment the transitive tree changes or resolves to `node-fetch@3` (ESM-only, where `require()` throws). Please add `node-fetch` explicitly to `dependencies`, pinned to `^2`.

src/commands/inspect.js
```diff
@@ -0,0 +1,57 @@
...
+    const mvnDir = path.join(process.cwd(), 'mvnw');
+    const jar = path.join(mvnDir, 'target', 'inspect.jar');
+    const server = spawn('java', [
+      '-cp',
+      jar,
+      'org.eolang.Inspect'
+    ], { cwd: mvnDir, stdio: ['pipe', 'pipe', 'pipe'] });
+    server.stdout.setEncoding('utf8').on('data', d => console.log(`[SERVER] ${d}`));
+    server.stderr.setEncoding('utf8').on('data', d => console.error(`[SERVER ERROR] ${d}`));
+    setTimeout(() => {
```
Favixx, line 22: Waiting a fixed `setTimeout(..., 2000)` for the server to come up is a race: on a slow/loaded machine the first request can fire before the port is bound (failure), and on a fast machine you always waste 2s. Better to poll the `/` endpoint until it answers (or read the server's stdout for a ready marker) and proceed on that signal.

src/commands/inspect.js
```diff
@@ -0,0 +1,57 @@
...
+const path = require('path');
+const os = require('os');
+const readline = require('readline');
+const fetch = require('node-fetch');
+
+module.exports = function(opts) {
+  return new Promise((resolve, reject) => {
+    const mvnDir = path.join(process.cwd(), 'mvnw');
+    const jar = path.join(mvnDir, 'target', 'inspect.jar');
+    const server = spawn('java', [
```
Favixx, line 15: No `server.on('error', ...)` handler. If `java` isn't on `PATH` or `inspect.jar` is missing, `spawn` emits an `error` event that is currently unhandled and will crash the process with an opaque stack. Since the surrounding Promise already exposes `reject`, wire the spawn error to `reject(err)` with a friendly message (e.g. "run `npx grunt` first to build inspect.jar").

mvnw/pom.xml
```diff
@@ -35,10 +35,15 @@
       <version>5.13.1</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest</artifactId>
```
Favixx, line 40: The new `hamcrest` dependency has no `<scope>`, so it defaults to `compile` and ends up on the runtime classpath. Hamcrest is a test-only assertion library — please add `<scope>test</scope>`.

mvnw/pom.xml
```diff
@@ -35,10 +35,15 @@
...
+      <version>3.0</version>
+    </dependency>
   </dependencies>
   <build>
     <finalName>eoc</finalName>
-    <sourceDirectory>${eo.generatedDir}</sourceDirectory>
+    <sourceDirectory>../inspect/src/main/java</sourceDirectory>
```
Favixx, line 46: Switching `<sourceDirectory>` from `${eo.generatedDir}` to `../inspect/src/main/java` repoints the **primary** compile root away from the generated EO sources, then re-adds the generated dir via build-helper `add-source`. This inverts the primary/secondary source roles and is fragile. Please confirm the generated EO sources still compile under every profile (not just the default `inspect` one), ideally with a comment explaining why the swap is needed.

test/commands/test_inspect.js
```diff
@@ -0,0 +1,44 @@
...
+    const inspect = proxyquire('../../src/commands/inspect', {
+      'child_process': { spawn },
+      'node-fetch': async (url, opts) => {
+        lastBody = opts.body;
+        return { ok: true, text: async () => opts.body };
+      },
+    });
+    const inspectPromise = inspect({ stdin: input, stdout: out });
+    input.write('hello\n');
+    await new Promise(res => setTimeout(res, 5000));
```
Favixx, line 36: Two hard `setTimeout(5000)` sleeps (10s total) make this test slow and flaky — it's implicitly coupled to the 2000ms startup delay inside `inspect.js`. Prefer event-driven waiting (resolve when the echoed response arrives) or `sinon` fake timers, so the test is deterministic and fast.

test/commands/test_inspect.js
```diff
@@ -0,0 +1,44 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2025 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+const assert = require('assert');
+const sinon = require('sinon');
+const stream = require('stream');
+const proxyquire = require('proxyquire');
+describe('inspect command (mocked)', () => {
+  it('should send text and receive echoed response, then exit cleanly', async () => {
```
Favixx, line 10: Only the happy path is covered. There's no test for the `catch` branch (fetch failure → "Request failed"), for the spawn-error path, or for the Java server itself. At minimum a fetch-failure case would guard the error handling this command depends on.

Favixx (changes requested):
Thanks for the work on the `inspect` entry point. The overall shape (command → Java HTTP server via takes → assembly jar) is reasonable, but I don't think it's mergeable yet — there are a few correctness/build issues plus test-quality gaps. Details inline; most are quick to fix.

Main blockers:
1. `package.json` declares `mocha` three times with conflicting versions — looks like an unresolved merge.
2. `node-fetch` is `require`d but is not declared in `dependencies` (only present transitively).
3. Server startup relies on a fixed `setTimeout(2000)`, which is racy.
4. `hamcrest` is added at compile scope instead of test.

Happy to re-review once these are addressed.

yegor256:

0crat:
@favixx Thanks for the review! You've earned +23 points for this: +12 as a basis; +15 for the 1591 hits-of-code that you reviewed; -5 for very few (8) comments.  Your running score is [+34](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #550, opened 2025-06-11
fix: warn does not throw now

ChabanovX:
The tinted-console helper maps console.warn() messages to a non-existent “orange” color, which causes a runtime error whenever console.warn() is used.

I replaced the unsupported “orange” color with “yellow” in the console tinting logic to avoid runtime errors on console.warn()


yegor256:
@ChabanovX thanks!

Issue #547, opened 2025-06-10
npx grunt does not work


ErnestMatskevich:
After running `npx grunt`  there are failings:
- compile
- dataize
- link
- test

0crat:
@ernestmatskevich 🐛 Awesome bug report! You've earned +12 points and helped elevate our project's quality. Keep your eyes peeled for anything that [seems off](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) in the repository - we [welcome](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) all [reports](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html). Your keen observations drive our success! 🚀

roman-shaposhnikov:
This issue is resolved by the upstream fix to [`objectionary/lints#624`](https://github.com/objectionary/lints/issues/624).

Verification today (2026-05-17):

- `lints#624` shipped in **lints `0.0.53`** (released 2025-06-22, see [rultor confirmation](https://github.com/objectionary/lints/issues/624#issuecomment-2994017591)).
- This repo currently pins `eo-version.txt` = **`0.61.0`**, and [`eo-maven-plugin 0.61.0`](https://github.com/objectionary/eo/blob/0.61.0/eo-maven-plugin/pom.xml) transitively depends on `lints 0.0.60` — well past the fix.
- The `grunt` GitHub Actions workflow on `master` is green today on Ubuntu 24.04, macOS 15 and Windows 2022 ([run 25988401392](https://github.com/objectionary/eoc/actions/runs/25988401392)) — all four originally-failing suites (`compile`, `dataize`, `link`, `test`) pass.

Closing as resolved.

PR #554, opened 2025-06-15
chore(548): fix usage of grunt

Dirakon:


Dirakon:
Pipelines seems to be currently unstable. Probably related to https://github.com/objectionary/eoc/issues/547

Issue #540, opened 2025-06-07
Replace `.then()` with `async/await` in JS command files for better readability


netimaaa:
In many files of the `src/commands` and `src/commands/js` directories, promise handling via `.then()` is used. This code can be made more readable by replacing it with `async/await`.

Example before:
```JavaScript
module.exports = function(opts) {
  return eo2jsw('transpile', {...opts, alone: true, project: 'project'}).then((r) => {
    console.info(`JS sources are generated in ${rel(path.resolve(opts.target, 'project'))}`);
    return r;
  });
};
```

Example after:
```JavaScript
module.exports = async function(opts) {
  const result = await eo2jsw('transpile', {...opts, alone: true, project: 'project'});
  console.info(`JS sources are generated in ${rel(path.resolve(opts.target, 'project'))}`);
  return result;
};
```

yegor256:
@netimaaa sounds like a good catch. Do you know how to configure ESLint to defend us against such suboptimal code constructs?

yegor256:
@netimaaa let's add this rule to our ESLint config

netimaaa:
@yegor256 To clarify the task: should I only add the `prefer-await-to-then` rule to ESLint config so that such constructs are highlighted, or should I also perform a mass replacement of all `.then()` calls with `async/await` in the code?

yegor256:
@netimaaa if you only add the rule the build won't pass and we won't be able to merge your changes to the master. Maybe you add the rule, fix it in a few files, and suppress the rule for other files. Then, in further pull requests, fix all files.

PR #555, opened 2025-06-17
Replace .then() with async/await (#540)

aequor-cntrl:


aequor-cntrl:
@0crat Hello, thank you! Should I remake the pull request with the proper branch name?

yegor256:
@aequor-cntrl leave it as is, but in the future try to use proper names.

PR #611, opened 2025-07-20
Add normalize command to process and rewrite EO programs

niotu:


yegor256:
@niotu Thanks for your contribution! Unfortunately, we can't merge it without tests. Every feature we add and every bug we fix must be accompanied by at least one new test that confirms that the code was not working before. Without such a test, how can we know that what you commit actually works?

yegor256:
@niotu what's up with this one? Would be great if you can continue and add tests.

yegor256:

Issue #637, opened 2025-08-21
brew tap should use specific tag, not master


yegor256:
Currently, we recommend doing this (in our README):

```
brew install objectionary/eoc
```

Maybe we can recommend something like this:

```
brew install objectionary/eoc@0.33.0
```

We need to point the user to a specific tag, not to unstable `master`.

vladislav-yevtushenko:
@yegor256  please take a look when you have a time https://github.com/objectionary/eoc/pull/643

vladislav-yevtushenko:
@yegor256 should we close it ?

yegor256:
@vladislav-yevtushenko thanks!

PR #643, opened 2025-08-27
#637: update homebrew formula

vladislav-yevtushenko:
On each release, this workflow runs a script to check for all released versions and creates a versioned Homebrew formula for each available version.

### DIFF
--- a/.github/homebrew_formula_up.sh
+++ b/.github/homebrew_formula_up.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: Copyright (c) 2022-2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+
+set -euo pipefail
+
+GITHUB_REPO="objectionary/eoc"
+FORMULA_DIR="Formula"
+
+GITHUB_TOKEN="${1:-}"
+TAG_VERSION="${2:-}"
+
+if [ -z "$GITHUB_TOKEN" ]; then
+  echo "Error: GitHub token must be provided as the first argument." >&2
+  exit 1
+fi
+
+if [ -z "$TAG_VERSION" ]; then
+  echo "Error: Tag version must be provided as the second argument." >&2
+  exit 1
+fi
+
+VERSION="${TAG_VERSION#v}"
+
+ORIG_FORMULA="$FORMULA_DIR/eolang.rb"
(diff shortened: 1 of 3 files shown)


yegor256:
@vladislav-yevtushenko can you please create an issue first, to explain the problem we need to solve. Now it's not clear what is the motivation for this PR.

vladislav-yevtushenko:
@yegor256 I explored the approach you suggested and learned how it works, but it doesn’t cover all the cases we need.

If you want to users to be able to install any version via Homebrew, you have to keep a formula file for each version (not just one formula) — there’s no way around it.

Here’s why my solution is more reliable:

**No broken installs**
Homebrew looks for a formula file for every version you try to install (when formulas are stored in the repo, not Homebrew core). If a formula is missing, Homebrew simply fails.
My script checks all releases and ensures there’s a formula file for each one, so users can always install any released version — no surprises.

vladislav-yevtushenko:
@yegor256 it also recommended way to have an upstream tap repository for managing formulas. For example: "objectionary/homebrew-eoc" (doc: https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap#upstream-taps).
But I guess it should be done by repo owner.

yegor256:
@vladislav-yevtushenko how about this instead:

```
brew install https://raw.githubusercontent.com/objectionary/eoc/0.32.0/Formula/eolang.rb
```

Could work?

vladislav-yevtushenko:
@yegor256 I am afraid no, "brew install" doesn't support that kind of installation from remote link

yegor256:
@vladislav-yevtushenko are you sure? ChatGPT says that it does. Also, check this file: https://github.com/objectionary/eoc/blob/0.32.0/Formula/eolang.rb It seems that the data inside is incorrect - we mention 0.31.0 version everywhere.

yegor256:
@vladislav-yevtushenko got it. Don't you think that it would be more simple and elegant, if the script would just take the latest file from the `Formula/` directory, change the version everywhere inside it (using `sed`) and create a new file. Currently, we have a pretty complicated script, which is hard to maintain. Just copying an existing file and changing the version would be much simpler.

yegor256:
@vladislav-yevtushenko looks perfect, thanks!

yegor256:
@vladislav-yevtushenko it seems that we forgot to update the README, to explain users how to use specific versions

PR #661, opened 2025-09-08
Apply an eslint async/await rule for all js files

### LINKED ISSUE #540: Replace `.then()` with `async/await` in JS command files for better readability

danilax86:
Closes #540

### DIFF
--- a/src/commands/lint.js
+++ b/src/commands/lint.js
@@ -28,7 +28,8 @@ module.exports = async function(opts) {
       return r;
     } catch (error) {
       throw new Error(
-        'There are errors and/or warnings; you may disable warnings via the --easy option'
+        'There are errors and/or warnings; you may disable warnings via the --easy option',
+        { cause: error }
       );
     }
   }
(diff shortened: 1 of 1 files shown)


yegor256:
@danilax86 try to sync your branch with master, it may work

yegor256:
@rultor merge

Issue #741, opened 2025-10-19
Simplifying eoc installation for users of the nix package manager


morphqdd:
As a user of the [NixOS](https://nixos.org) operating system, I encountered a problem installing eoc via npm. This is due to how the nix package manager stores packages, which makes global installation via npm impossible. 

Attempting to use [bun](https://bun.com) instead of npm also did not produce the desired result. Since bun was unable to pull in the necessary dependencies, the basic example from README.md did not compile.

I want to write a flake.nix file that will simplify the installation of eoc for users of both NixOS and the Nix package manager in general. I also want to add a new installation method to README.md.

yegor256:
@morphqdd good point, please do it!

0crat:
@morphqdd **Congratulations!** 🎉 Thanks for reporting a new bug! According to our work policy, you've earned **+12 points** for helping improve our project quality. Your contribution is valuable and directly rewarded. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) it – each accepted bug report earns you another **12 points**!

morphqdd:
@yegor256 fix([#741](https://github.com/objectionary/eoc/pull/743))

morphqdd:
@yegor256 Sorry, Yegor, for the mess in the issue. I didn't know how to correctly reference a specific issue in the commit and how it would appear in GitHub.

PR #743, opened 2025-10-22
fix(#741): Added the ability to install the eoc package using flake.nix

morphqdd:
This pr adds `flake.nix` for installing `eoc` via the nix package manager. 

It also adds `flake.lock` to freeze the versions of nix tools used for installation and updates `REUSE.toml` to ignore the presence of license headers in `flake.lock`, as it is generated automatically and contains only meta information for building, similar to `package.lock.json`. 

`README.md` has also been updated, and a new way to download via nix has been added.


morphqdd:
@yegor256 take a look, please

PR #771, opened 2025-11-06
fix(#767): codecov-action action@v5 changes its file to files

Thayorns:
@yegor256 
This PR covers #767 
Since that `codecov-action@v5` changed it's `file` parameter to `files`, it should pass


Thayorns:
@yegor256, ready for review

yegor256:
@Thayorns thanks!

Issue #778, opened 2025-11-13
`prints Maven own version` test should use async/await


Thayorns:
@yegor256 

This test [prints Maven own version](https://github.com/objectionary/eoc/blob/master/test/test_mvnw.js) 
```
it('prints Maven own version', (done) => {
    const opts = {batch: true};
    mvnw(['--version', '--quiet'], null, opts.batch);
    done();
  });
```

should be written with `async/await` because `mvnw()` function is asynchronous. The test calls `done()` synchronously and always passes - even if the asynchronous process fails.

PR #780, opened 2025-11-13
fix(#778): prints Maven own version test uses async/await

Thayorns:
@yegor256 
This PR covers #778 issue and refactors the test to use `async/await` syntax, so when the code passes the `await` line, the test passes successfully.


yegor256:
@Thayorns submit a bug about a flaky test

PR #931, opened 2026-05-05
Document missing pipeline commands in README (#857)

### LINKED ISSUE #857: automate generation of `Commands` and `Options` in the README
Currently, we have [Commands](https://github.com/objectionary/eoc?tab=readme-ov-file#commands) section in the README. It's manually written. That said, it's some sort of duplication, because we can get the same information using `--help` option. As it was mentioned [recently](https://github.com/objectionary/eoc/pull/850#discussion_r2815210762):

> What happens when the actual list of options of eoc changes and we forget to update the README file?

Thus, it's better to generate this information automatically. We also have an option to remove this section completely, but sometimes it's quite useful to check that the tool can do without running it.

bibonix:
@yegor256 ready for merge — all 16 CI checks are green.

Closes #857.

The README's Commands section was hand-written and had drifted from the actual CLI: `clean`, `parse` and `resolve` are exposed by `eoc --help` but were not mentioned in `README.md`. As discussed in the issue, the goal is to keep the documentation in sync with the CLI automatically rather than relying on contributors to remember to update both.

This PR takes the test-driven approach you typically prefer here:

* `test/test_readme.js` runs `eoc --help`, parses its `Commands:` block, and asserts every command name (except the built-in `help`) appears in `README.md` between backticks. On `master` the test fails with `clean, parse, resolve` reported as missing.
* `README.md` is updated to document the three missing commands and `lint` is moved next to `assemble` so the bullet order matches the pipeline order in `src/eoc.js`.

Whenever a new `program.command(...)` is added in `src/eoc.js` (or one is renamed/removed) the test fails until the README catches up — same kind of CI guard #857 was asking for, without the editorial loss of switching to a fully generated section.

Two-commit history:
1. `test(#857): assert README mentions every command from eoc --help` — failing test that reproduces the drift.
2. `fix(#857): document the missing pipeline commands in README` — README change that turns the test green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,8 +116,11 @@ You can also do many other things with `eoc` commands
 (the flow is explained in [this blog post][blog]):
 
 * `register` finds necessary `.eo` files and registers them in a JSON catalog
+* `parse` parses `.eo` files into `.xmir` without pulling foreign objects
 * `assemble` parses `.eo` files into `.xmir`, optimizes them,
   and pulls foreign EO objects
+* `lint` finds style-related errors in EO and XMIR files
+* `resolve` resolves all the dependencies required for compilation
 * `transpile` converts `.xmir` files to the target programming
 language (Java by default)
 * `compile` converts target language sources (e.g., `.java`)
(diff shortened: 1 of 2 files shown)


0crat:
@bibonix Hey there! 🎉 Nice work on that contribution - you've snagged **+16 points** which puts your running score at [+192](https://www.eolang.org/zerocracy/objectionary-vitals.html)! The bonus calculation considered your hits-of-code, review feedback, and overall contribution quality according to our policy. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the solid contributions rolling in, and don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) to track your progress!

Issue #892, opened 2026-04-19
Outdated 'sodg' Command Needs Removal


yegor256:
It's outdated and must be removed.

yegor256:
@rultor release, tag is `0.35.0`

yegor256:
@rultor release, tag is `0.35.0`

yegor256:
@rultor release, tag is `0.35.0`

yegor256:
@rultor release, tag is `0.35.0`

yegor256:
@rultor release, tag is `0.35.0`

0crat:
@yegor256 Thanks for closing this issue! You've earned **+8 points** following our bonus policy: **+8** base points, **+0** for quick resolution (took more than 12 hours), and **-0** penalty for 1 day of delay (within acceptable limits). Your running score is [+235](https://www.eolang.org/zerocracy/objectionary-vitals.html) - keep up the great work! Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates. 🎯

PR #949, opened 2026-05-20
#939: drop dead postinstall.js and stale patch

### LINKED ISSUE #939: `postinstall.js` skips `patch-package` due to outdated `grunt-mocha-cli` dependency check

nacNAC333:
The `grunt-mocha-cli` devDependency was replaced by `grunt-mocha-test`
in #892, but `scripts/postinstall.js` still checked for
`node_modules/grunt-mocha-cli` to decide whether to run
`patch-package`.  That check was always false, so the Windows
shell-fix patch in `patches/grunt-mocha-cli+7.0.0.patch` was
never applied.  `grunt-mocha-test` runs mocha in-process and
does not need the patch.

Removed:
- `scripts/postinstall.js` (dead conditional logic)
- `patches/grunt-mocha-cli+7.0.0.patch` (targets absent package)
- `patch-package` devDependency (no remaining patches)
- `postinstall` npm script entry

Closes #939.


nacNAC333:
Parent issue #939 was resolved and closed. Closing this PR.

PR #955, opened 2026-05-22
#94: Add timestamps for maven logs

### LINKED ISSUE #94: Add timestamps for maven logs
I was struggling with the [task](https://github.com/objectionary/eo/issues/1428) of adding timestamps for build logs, because applied changes for `log4j` logger didn’t applied on a build process. In other words, if you use `eoc` with the new version of `eo-maven-plugin` (where we already added timestamps for `log4j` logs), you still will see the next output:

```
[INFO] There are 33 EO sources registered already
[INFO] Registered 1 EO sources from ./Users/lombrozo/Workspace/EOlang/Projects/learning to ./Users/lombrozo/Workspace/EOlang/Projects/learning/.eoc/eo-foreign.json, included [**.eo], excluded [.eoc/**]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  0.536 s
[INFO] Finished at: 2022-11-10T21:46:32+03:00
[INFO] ------------------------------------------------------------------------
```

Why? Because we use logs frameworks like this (I can be wrong in details):

￼￼￼
image

As you can see, we expect to use `log4j` implementation for `slf4j`, but here is a trick: Since 3.1.x version of maven, the default maven distributive uses  its own `slf4j` implementation - `slf4j-simple`, in other words, our picture changes to the next:

￼
image

You can check maven that installed on your laptop and in the maven `lib` directory you probably will find the next jars:
```
maven-slf4j-provider-3.8.3.jar (==“slf4j-simple”)
slf4j-api-1.7.32.jar
```

Since `slf4j` chooses only one logging implementation, we have only 2 options here to add timestamps for build logs:
1. Use native `slf4j-simple` logger directly. To do so we can just add the next arguments for all maven commands  
```
-Dorg.slf4j.simpleLogger.showDateTime=true
-Dorg.slf4j.simpleLogger.dateTimeFormat=yyyy-dd-MM HH:mm:ss
``` 
Then all maven build logs will look like:

```
2022-10-11 21:49:00 [INFO] Assemble cycle #1 (eo:33/xmir:33/xmir2:33/discovered:33 -> eo:33/xmir:33/xmir2:33/discovered:33), took 3s
2022-10-11 21:49:00 [INFO] 1 assemble cycle(s) produced some new object(s): eo:33/xmir:33/xmir2:33/discovered:33
2022-10-11 21:49:00 [INFO] ------------------------------------------------------------------------
2022-10-11 21:49:00 [INFO] BUILD SUCCESS
2022-10-11 21:49:00 [INFO] ------------------------------------------------------------------------
2022-10-11 21:49:00 [INFO] Total time:  3.203 s
2022-10-11 21:49:00 [INFO] Finished at: 2022-11-10T21:49:00+03:00
2022-10-11 21:49:00 [INFO] ------------------------------------------------------------------------
```

2. Remove `slf3j-simple` from `mvnw` distributive (I don’t know if it’s even possible). But at least, you can try to remove `slf4j-simple` from your `maven/lib` folder and `log4j` will work fine.

___
Useful links:
1. More detailed about maven logging you can read [here](https://maven.apache.org/maven-logging.html).
2. Maven [slf4j provider library](https://repo.maven.apache.org/maven2/org/apache/maven/maven-slf4j-provider/3.8.3/) in [GitHub](https://github.com/apache/maven/tree/master/maven-slf4j-provider).
3. [The list of properties](https://github.com/apache/maven/blob/master/apache-maven/src/assembly/maven/conf/logging/simplelogger.properties) that we can use in order to adjust maven `slf4j-simple` logger.

nacNAC333:
Fixes #94

Add timestamps to Maven log output for easier debugging.

Changes to src/mvnw.js:
- Add timestamp() function for [HH:MM:SS] formatted log prefixes
- Pipe Maven stdout/stderr to add timestamps
- Makes debugging long builds much easier

### DIFF
--- a/src/mvnw.js
+++ b/src/mvnw.js
@@ -21,6 +21,18 @@ function shell() {
   }
 }
 
+/**
+ * Format a timestamp for log lines.
+ * @return {String} Timestamp string in [HH:MM:SS] format
+ */
+function timestamp() {
+  const now = new Date();
+  const h = String(now.getHours()).padStart(2, '0');
+  const m = String(now.getMinutes()).padStart(2, '0');
+  const s = String(now.getSeconds()).padStart(2, '0');
+  return `[${h}:${m}:${s}]`;
+}
+
 let beginning,
   phase = 'unknown',
   running = false,
@@ -99,10 +111,28 @@ module.exports.mvnw = function(args, tgt, batch) {
       process.platform === 'win32' ? params.map((p) => `"${p}"`) : params,
       {
         cwd: home,
-        stdio: 'inherit',
+        stdio: ['pipe', 'pipe', 'inherit'],
         shell: shell(),
       }
     );
+    result.stdout.on('data', (chunk) => {
+      const ts = timestamp();
+      const lines = chunk.toString().split('\n');
+      for (const line of lines) {
+        if (line.trim()) {
+          process.stdout.write(`${ts} ${line.trim()}\n`);
+        }
+      }
+    });
+    result.stderr.on('data', (chunk) => {
+      const ts = timestamp();
+      const lines = chunk.toString().split('\n');
+      for (const line of lines) {
+        if (line.trim()) {
+          process.stderr.write(`${ts} ${line.trim()}\n`);
+        }
+      }
+    });
     if (tgt !== undefined && args.includes('--quiet')) {
       if (!batch) {
         start();
--- a/src/mvnw.js
+++ b/src/mvnw.js
@@ -113,33 +113,44 @@ module.exports.mvnw = function(args, tgt, batch) {
       process.platform === 'win32' ? params.map((p) => `"${p}"`) : params,
       {
         cwd: home,
-        stdio: ['pipe', 'pipe', 'inherit'],
+        stdio: ['pipe', 'pipe', 'pipe'],
         shell: shell(),
       }
     );
+    let outBuf = '';
+    let errBuf = '';
     result.stdout.on('data', (chunk) => {
-      const ts = timestamp();
-      const lines = chunk.toString().split('\n');
-      for (const line of lines) {
-        if (line.trim()) {
-          process.stdout.write(`${ts} ${line.trim()}\n`);
-        }
+      outBuf += chunk.toString();
+      let idx;
+      while ((idx = outBuf.indexOf('\n')) !== -1) {
+        process.stdout.write(`${timestamp()} ${outBuf.slice(0, idx)}\n`);
+        outBuf = outBuf.slice(idx + 1);
       }
     });
     result.stderr.on('data', (chunk) => {
-      const ts = timestamp();
-      const lines = chunk.toString().split('\n');
-      for (const line of lines) {
-        if (line.trim()) {
-          process.stderr.write(`${ts} ${line.trim()}\n`);
-        }
+      errBuf += chunk.toString();
+      let idx;
+      while ((idx = errBuf.indexOf('\n')) !== -1) {
+        process.stderr.write(`${timestamp()} ${errBuf.slice(0, idx)}\n`);
+        errBuf = errBuf.slice(idx + 1);
       }
     });
+    function flush() {
+      if (outBuf) {
+        process.stdout.write(`${timestamp()} ${outBuf}\n`);
+        outBuf = '';
+      }
+      if (errBuf) {
+        process.stderr.write(`${timestamp()} ${errBuf}\n`);
+        errBuf = '';
+      }
+    }
     if (tgt !== undefined && args.includes('--quiet')) {
       if (!batch) {
         start();
       }
       result.on('close', (code) => {
+        flush();
         if (code !== 0) {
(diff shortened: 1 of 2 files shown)


yegor256:
@nacNAC333  Thanks for your contribution! Unfortunately, we can't merge it without tests. Every feature we add and every bug we fix must be accompanied by at least one new test that confirms that the code was not working before. Without such a test, how can we know that what you commit actually works?

src/mvnw.js
```diff
@@ -99,10 +111,28 @@ module.exports.mvnw = function(args, tgt, batch) {
       process.platform === 'win32' ? params.map((p) => `"${p}"`) : params,
       {
         cwd: home,
-        stdio: 'inherit',
+        stdio: ['pipe', 'pipe', 'inherit'],
```
Favixx: `['pipe', 'pipe', 'inherit']` means stdin=pipe, stdout=pipe, **stderr=inherit**. With stderr inherited, `child.stderr` is `null`, which makes the `result.stderr.on(...)` handler below throw. Two separate problems here:

- To pipe both stdout and stderr (so you can prefix them), use `['inherit', 'pipe', 'pipe']` and attach handlers to both.
- Changing stdin from the previous `inherit` to `pipe` also detaches the child from the terminal, which can break interactive prompts.

src/mvnw.js
```diff
@@ -99,10 +111,28 @@ module.exports.mvnw = function(args, tgt, batch) {
...
+    result.stdout.on('data', (chunk) => {
+      const ts = timestamp();
+      const lines = chunk.toString().split('\n');
+      for (const line of lines) {
+        if (line.trim()) {
+          process.stdout.write(`${ts} ${line.trim()}\n`);
+        }
+      }
+    });
+    result.stderr.on('data', (chunk) => {
```
Favixx, line 130: **Crash site.** Because stderr is inherited above, `result.stderr` is `null`, so this line throws `TypeError: Cannot read properties of null (reading 'on')` synchronously, before Maven runs. Verified with `mvnw(['--version'])` on this branch:

```
TypeError: Cannot read properties of null (reading 'on')
    at src/mvnw.js:127:19
```

Either pipe stderr in the `stdio` array, or drop this handler if stderr is meant to stay inherited.

src/mvnw.js
```diff
@@ -99,10 +111,28 @@ module.exports.mvnw = function(args, tgt, batch) {
       process.platform === 'win32' ? params.map((p) => `"${p}"`) : params,
       {
         cwd: home,
-        stdio: 'inherit',
+        stdio: ['pipe', 'pipe', 'inherit'],
         shell: shell(),
       }
     );
+    result.stdout.on('data', (chunk) => {
```
Favixx, line 122: Two issues with the line-by-line reformatting:

- `line.trim()` strips Maven's indentation and collapses whitespace, so structured output (e.g. nested reactor logs) is mangled.
- A `data` chunk is not guaranteed to align to line boundaries — a chunk can split a line mid-way, so `split('
')` per chunk will timestamp partial lines and concatenate them incorrectly. Buffer until a newline (e.g. `readline.createInterface`) if you keep this approach.

Favixx (changes requested):
**`mvnw` crashes on every invocation — `result.stderr` is `null`**

This change makes the `mvnw` wrapper throw a `TypeError` on every call, so effectively every `eoc` command (parse, compile, etc.) fails before Maven even starts.

Setting `stdio: ['pipe', 'pipe', 'inherit']` inherits the child's stderr, which means `child.stderr` is `null`. The handler added below then dereferences it.

Reproduced on this branch with a minimal call `mvnw(['--version'])`:

```
TypeError: Cannot read properties of null (reading 'on')
    at src/mvnw.js:127:19
    at new Promise (<anonymous>)
```

The same call on `master` runs to completion, so the regression is introduced here.

Additional notes:
1. Reformatting Maven output line-by-line in JS is lossy (see inline comments). A native approach via slf4j-simple flags (`-Dorg.slf4j.simpleLogger.showDateTime=true` / `-Dorg.slf4j.simpleLogger.dateTimeFormat=...`), as in #964/#909, lets Maven emit timestamps itself and preserves formatting.
2. Per project policy, please add a test that fails before this change and passes after — none of the current tests exercise the spawn path, which is why this crash was not caught.

nacNAC333:
@yegor256 Test added — `prefixes mvnw output lines with [HH:MM:SS] timestamp` captures stdout from mvnw and verifies the `[HH:MM:SS]` prefix appears on piped output lines. Rebased on latest master, all 6 mvnw tests pass.

yegor256:

0crat:
@favixx Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+11](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #956, opened 2026-05-23
fix: remove stale patch-package postinstall

### LINKED ISSUE #939: `postinstall.js` skips `patch-package` due to outdated `grunt-mocha-cli` dependency check

999axel999:
## Summary
- Remove the stale `patch-package`/`grunt-mocha-cli` postinstall hook and its unused devDependency.
- Keep a no-op `scripts/postinstall.js` with a `#939` TODO documenting the follow-up Windows-shell investigation instead of silently deleting the context.
- Bump the bundled EO parser version from `0.61.0` to `0.61.1`, which fixes the parser-side failure that made the earlier cleanup look suspicious.
- Add a regression test proving the stale install hook and lockfile metadata are gone.

Closes #939

## Tests
- `node` parser smoke check for slash syntax against EO `0.61.1`
- `.\node_modules\.bin\mocha.cmd test/test_postinstall.js --timeout 1200000`
- `npm.cmd exec -- eslint test/test_postinstall.js scripts/postinstall.js`
- `git diff --check`
- GitHub Actions for `7f3e79777a4f61ad85e5a1901cac42bd93bca6b1`: `pdd`, `markdown-lint`, `copyrights`, `yamllint`, `shellcheck`, `typos`, `actionlint`, `eslint`, `reuse`, `xcop`, `itest`, and `grunt` all passed.


yegor256:
@999axel999 please, fix merge conflicts

yegor256:
@999axel999 this change looks weird: you don't change anything, just deleting what works. Apparently, CI is clean after your changes, but this most probably means that CI is weak.

999axel999:
@yegor256 I reworked the PR after your feedback. It no longer relies on a blind deletion: the branch now bumps the EO parser from `0.61.0` to `0.61.1`, keeps a no-op `scripts/postinstall.js` with a `#939` follow-up TODO explaining the Windows-shell patch question, and adds a regression test that verifies the stale `patch-package` hook and lockfile metadata are gone. The current head `7f3e797` is green in GitHub Actions, including `itest` and `grunt`.

yegor256:
@999axel999 fixed already

PR #958, opened 2026-05-26
#939: remove obsolete postinstall patch

### LINKED ISSUE #939: `postinstall.js` skips `patch-package` due to outdated `grunt-mocha-cli` dependency check

akmhatey-ai:
Fixes #939.

This removes the obsolete `patch-package` postinstall path left behind after the project switched from `grunt-mocha-cli` to `grunt-mocha-test` in #892. The stale script checked for `node_modules/grunt-mocha-cli`, which is no longer installed, and the only patch under `patches/` also targeted that removed package. Since `grunt-mocha-test` runs Mocha through its library wrapper rather than the old spawned CLI path, there is no equivalent patch target to keep.

Changes:
- remove `scripts/postinstall.js`
- remove `patches/grunt-mocha-cli+7.0.0.patch`
- remove `patch-package` from dev dependencies
- remove the `postinstall` package script

Validation:
- `npm ci` succeeds and no longer prints `Production environment detected, skipping patch-package.`
- `npx grunt --file=test/test_eoc.js` -> 8 passing
- `npx mocha test/**/*.js --timeout 1200000 --grep eoc` -> 9 passing
- `npx mocha test/test_eoc.js --timeout 1200000` -> 8 passing
- `npx eslint .`
- `git diff --check`
- `gitleaks detect --source . --no-banner --redact --log-opts HEAD`

Limitations:
- Full `npx grunt` was attempted locally, but this Windows environment does not have `JAVA_HOME`/`javac` configured, so Java-backed EO/Maven tests fail before exercising this package cleanup.


akmhatey-ai:
Closing this as duplicate. After opening it I rechecked the current PR queue and confirmed that #949 and #956 already make the same cleanup: remove the obsolete postinstall script, remove the stale grunt-mocha-cli patch, drop patch-package, and update package-lock.json. This PR does not add distinct value.

PR #968, opened 2026-06-03
fix(#966): format relative sources in place

### LINKED ISSUE #966: `eoc fmt` command doesn't overwrite sources in place when `--sources` is a relative path
### What happens

The `fmt` command is documented as "Format EO files in the source directory" and is expected to overwrite the `.eo` sources **in place**. However, when `--sources` is a **relative** path (including the default `.`), the formatted files are written under `--target` instead, and the original sources are left untouched — silently, with `BUILD SUCCESS` and exit code `0`.

### How to reproduce

Verified locally with JDK 21 + Maven 3.9.16, parser `0.61.0`, home-tag `0.59.7`.

`project/src/app.eo`:
```
# EO file header
# Copyright notice for testing
# This is a comment for the app
[] > app
```

Run with a **relative** `-s` from the project directory:
```
cd project
eoc fmt -s src -t target --parser 0.61.0 --home-tag 0.59.7
```

Observed:
- `src/app.eo` is **unchanged** (the source is not formatted).
- The formatted result appears at **`target/src/app.eo`** (`# No comments.` / `[] > app`).
- The log itself prints: `XMIR files converted to EO files at target\src`.

Running the same command with an **absolute** `-s` formats the source in place correctly, which is why the issue is easy to miss.

### Root cause

`fmt` (`src/eoc.js`) passes the sources path as `printOutput`:
```js
await coms().print({
  printInput: '1-parse',
  printOutput: program.opts().sources,
  ...program.opts()
});
```

but `print.js` resolves `printOutput` relative to `--target`:
```js
const output = path.resolve(opts.target, opts.printOutput); // src/commands/print.js:19
```

`path.resolve(target, sources)` equals `sources` only when `sources` is **absolute** (the first argument is ignored). For a **relative** `sources` it yields `<target>/<sources>`, which is the wrong directory.

| `--sources` | `path.resolve(target, sources)` | result |
| --- | --- | --- |
| absolute | first arg ignored → the absolute path | correct (source dir) |
| `src` | `target/src` | wrong |
| `.` (default) | `target` (i.e. `.eoc`) | wrong |

This is correct for the `print` **command** itself, whose `--print-input`/`--print-output` are documented as "relative to --target". The bug is that `fmt` reuses `print.js` with a path that is relative to **CWD**, not to `--target`.

### Test coverage gap (why this went unnoticed)

`test/commands/test_fmt.js` always passes an **absolute** `-s`:
```js
const source = path.resolve(home, 'src')   // always absolute
fmt(source, target)
```

Because `sources` is absolute, `path.resolve(target, source) === source`, so the test only ever exercises the one case where the bug is hidden. It stays green even though the **default and most common real-world invocation** (`eoc fmt` with the default `-s .`, or `eoc fmt -s src`) is broken. The suite never covers a relative `--sources`, which is exactly the scenario users hit.

### Confirmation from a sibling command

The `normalize` command performs the same "write EO back to the sources" operation and does it correctly — it resolves sources to absolute first:
```js
const sources = path.resolve(opts.sources);   // src/commands/normalize.js:33
...
`-Deo.printOutputDir=${sources}`
```
This confirms the intended behavior is to write into the actual sources directory, and that `fmt` deviates from it.

### Suggested fix

Resolve the sources path to absolute before handing it to `print`, e.g. in the `fmt` action:
```js
printOutput: path.resolve(program.opts().sources),
```
and extend `test/commands/test_fmt.js` to also run with a **relative** `--sources` so the regression is covered.

nanookclaw:
## What changed

`eoc fmt` now passes the source directory to `print` as an absolute path. This keeps the existing standalone `print` behavior, where `--print-output` is relative to `--target`, but prevents the `fmt` wrapper from treating a relative `--sources` value as a target-relative output directory.

I also added a regression test for relative `--sources`: it verifies the original source file is formatted in place and that no mirrored output is created under `--target/<relative-sources>`.

Closes #966.

## Verification

- `npm ci`
- `npx mocha test/commands/test_fmt.js --timeout 1200000`
- `npx eslint src/eoc.js test/commands/test_fmt.js`
- `node --check src/eoc.js`
- `node --check test/commands/test_fmt.js`
- `git diff --check`

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -372,7 +372,7 @@ program.command('fmt')
     await pipe()(coms(), ['register', 'parse'], program.opts());
     await coms().print({
       printInput: '1-parse',
-      printOutput: program.opts().sources,
+      printOutput: path.resolve(program.opts().sources),
       ...program.opts()
     });
   });
--- a/test/commands/test_fmt.js
+++ b/test/commands/test_fmt.js
@@ -148,4 +148,20 @@ describe('fmt', () => {
     })
     done()
   })
+  it('formats relative sources in place without writing under target', done => {
+    const [source, target] = prepareTestDirectory()
+    const file = createFile(source, testCases[0].before)
+    const relativeSource = path.relative(process.cwd(), source)
+    fmt(relativeSource, target)
+    assert.strictEqual(
+      fs.readFileSync(file, 'utf8'),
+      testCases[0].after,
+      'Relative --sources must be overwritten in place'
+    )
+    assert(
+      !fs.existsSync(path.resolve(target, relativeSource, 'app.eo')),
+      'Relative --sources must not create formatted files under --target'
+    )
+    done()
+  })
 })


0crat:
@nanookclaw Hello! I noticed that your Git branch name "`fix/fmt-relative-sources`" doesn't follow our [naming convention](https://www.yegor256.com/2014/04/15/github-guidelines.html). According to our team policy, this results in a **-6 point** deduction. For future branches, please use the ticket number as the branch name - in this case, "`967`" would be the correct format. Your current running score is [-6](https://www.eolang.org/zerocracy/objectionary-vitals.html), and you can check your [Zerocracy account](https://www.zerocracy.com/) for updates.

0crat:
@nanookclaw Thanks for the contribution! You've earned **+12 points** for this: **+16** as a basis per our policy, but **-4** for contributing fewer than 40 hits-of-code (you had 18). According to our [work policy](https://www.yegor256.com/2014/11/14/hits-of-code.html), contributions under 40 hits-of-code receive this deduction to encourage more substantial contributions. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming with larger changesets when possible! Your running score is [+6](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #976, opened 2026-06-05
Remove duplicate invalid parser-version test in test_parse.js

### LINKED ISSUE #972: Duplicate invalid-version tests in `test_parse.js` do not cover separate behaviors
The tests `rejects invalid parser version` at `test/commands/test_parse.js:42` and `validates parser version before calling Maven` at line 69 are the same test twice. Both create a project, run `parse --parser=999.999.999`, and wrap it in `assert.throws`; the only differences are the temp directory name and the string handed to `assert.throws`.

The problem this creates is unprotected behavior hiding behind a passing test. The second name points at the fast-fail guard in `src/mvnw.js:45`, which calls `process.exit(1)` before the `mvnw` subprocess is ever spawned. The body never checks that, though, it only checks that something throws. A bogus `--parser` also makes Maven itself fail late when it cannot fetch the matching pom, and `assert.throws(fn, 'string')` treats the string as the no-throw message rather than an error matcher, so the test cannot tell an early exit apart from a late Maven failure. Remove the early guard and this test stays green, so the very behavior it is named for is not covered.

Either drop one of the two cases, or make `validates parser version before calling Maven` assert a distinguishing signal, such as the early `is not available in Maven Central` message in the output or the absence of Maven artifacts under `target`.

Thayorns:
The cases `rejects invalid parser version` and `validates parser version before calling Maven` ran the same command (`parse --parser=999.999.999`) with the same `assert.throws`, differing only in the temp-directory name and the string handed to `assert.throws`. That string was read as the no-throw message rather than a matcher, so neither case checked how the command actually failed. This drops the duplicate and makes the surviving case verify a real signal: the run must exit with status 1.

In place of the duplicate I added a separate case for a malformed version (`--parser=not-a-version`), so the suite now covers two kinds of bad input instead of the same one twice. I did look at making the second case assert that no Maven artifacts land under `target`, to prove the early guard in `src/mvnw.js` rejects the version before Maven runs. That does not work: a bogus version that reaches Maven also fails with exit 1 and leaves no `target`, so the absence of artifacts distinguishes nothing. The only thing that separates the early guard from a late Maven failure is the console line, which we do not assert on, so removing the duplicate is the right call.

Closes #972

### DIFF
--- a/test/commands/test_parse.js
+++ b/test/commands/test_parse.js
@@ -44,15 +44,14 @@ describe('parse', () => {
     const home = path.resolve('temp/test-parse/invalid-version');
     createTestProject(home);
     assert.throws(
-      () => {
-        runSync([
-          'parse',
-          '--parser=999.999.999',
-          '-s', path.resolve(home, 'src'),
-          '-t', path.resolve(home, 'target'),
-        ]);
-      },
-      'Command should fail with invalid parser version'
+      () => runSync([
+        'parse',
+        '--parser=999.999.999',
+        '-s', path.resolve(home, 'src'),
+        '-t', path.resolve(home, 'target'),
+      ]),
+      (error) => error.status === 1,
+      'invalid parser version was not rejected with a non-zero exit'
     );
   });
   it('accepts valid parser version', function () {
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 the failing `itest` check here is not caused by this PR. This change only touches `test/commands/test_parse.js` (a unit test), which the `itest` workflow never runs.

`master` already fails the same workflow with the identical errors: the latest run on `master` (commit #968) has the same 25 `token recognition error at: '/'` failures on the pulled objects (`number.eo`, `bytes.eo`, `malloc.eo`, `sm/posix.eo`, ...), while every itest run on 2026-06-04 and earlier was green. This branch is just inheriting that breakage.

The root cause looks like a version mismatch: `eo-version.txt` is `0.61.0` (the parser plugin) but `home-tag.txt` is `0.59.7`, and the pulled objects use the typed-tail syntax (`[] > as-i64 /i64`, `[x] > gt /true`) that the lexer in `0.61.0` does not recognize, so `eo:transpile` aborts.

Locally the unit suite for this change is green and eslint is clean. Happy to adjust naming or split things differently if you'd like.

yegor256:
@Thayorns see: https://www.yegor256.com/2025/04/19/dont-merge-into-broken-master.html

Thayorns:
@yegor256 take a look please, CI is green now

Thayorns:
@yegor256 review please

0crat:
@thayorns Thanks for the contribution! You've earned **+16 points** according to our team policy for providing a code review. Your running score is [+160](https://www.eolang.org/zerocracy/objectionary-vitals.html). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too!

PR #977, opened 2026-06-06
fix: declare test helper variables

### LINKED ISSUE #969: Undeclared variables in `test/commands/java/test_pipeline.js` causing potential global scope conflict
The tests in `test/commands/java/test_pipeline.js` assign their helper variables without a declaration: `calls = []` at lines 15, 48 and 66, and `opts = {...}` at lines 19 and 52. The file is plain CommonJS with no `'use strict'`, so each bare assignment creates a property on `globalThis` instead of a local. The same happens to `numberOfPlaceholders` in the two `for` loops of `test/commands/test_generate_comments.js` at lines 26 and 43.

The result is that the three pipeline tests share one mutable variable through the global object rather than staying isolated. They pass today only because each test resets `calls = []` at its top; a test that forgets the reset inherits the previous test's array. That makes the order of tests significant and turns any future name clash on `globalThis` into a silent corruption. It is also a latent crash, because the moment the file gains `'use strict'` or becomes an ES module every one of these assignments throws `ReferenceError`.

The leak is easy to see by adding a test that never touches `calls` and asking whether the global is clean:

```js
it('does not leak the maven accumulator into global scope between tests', () => {
  assert.strictEqual(globalThis.calls, undefined, 'calls leaked onto the global object');
});
```

Against the current code it fails, and the diff shows the global still holding the previous test's command list instead of nothing:

```
AssertionError [ERR_ASSERTION]: calls leaked onto the global object
+ actual - expected
+ [ [ 'eo:resolve', 'eo:place', '-Deo.version=undefined', '-Deo.tag=undefined', ... ] ]
- undefined
```

Declaring the variables fixes it without any change in behavior, since each is already assigned fresh on every run: `const calls = []` and `const opts = {...}` inside each test in `test/commands/java/test_pipeline.js`, and `let numberOfPlaceholders` in the two loops of `test/commands/test_generate_comments.js`.

999axel999:
## Summary
- Declare `calls` and `opts` inside `test/commands/java/test_pipeline.js` tests instead of leaking globals.
- Declare `numberOfPlaceholders` inside the `generate_comments` loops.
- Add regression checks proving the helpers no longer land on `globalThis`.

Closes #969

## Tests
- `node ..\objectionary__eoc_fix\node_modules\mocha\bin\mocha.js test\commands\java\test_pipeline.js --timeout 1200000`
- `node ..\objectionary__eoc_fix\node_modules\mocha\bin\mocha.js test\commands\test_generate_comments.js --grep "does not leak placeholder" --timeout 1200000`
- `node ..\objectionary__eoc_fix\node_modules\eslint\bin\eslint.js test\commands\java\test_pipeline.js test\commands\test_generate_comments.js`
- `git diff --check origin/master...HEAD`

Note: full `test/commands/test_generate_comments.js` is not runnable in this local Windows shell because its helper spawns bare `node` via `execSync`, and that child command is not found in this environment. The new leak regression itself was run and passes.

### DIFF
--- a/test/commands/test_generate_comments.js
+++ b/test/commands/test_generate_comments.js
@@ -23,7 +23,7 @@ describe('generate_comments', () => {
   });
   it('fills output depending on the number of placeholders in the input code', (done) => {
     const home = makeHome();
-    for (numberOfPlaceholders = 0; numberOfPlaceholders < 3; ++numberOfPlaceholders) {
+    for (let numberOfPlaceholders = 0; numberOfPlaceholders < 3; ++numberOfPlaceholders) {
       const exampleInput =
         '# <COMMENT-TO-BE-ADDED>\n'.repeat(numberOfPlaceholders);
       const outputFilePath = path.resolve(home, 'out.json');
@@ -40,7 +40,7 @@ describe('generate_comments', () => {
   });
   it('fills output as expected when encountering valid EO code', (done) => {
     const home = makeHome();
-    for (numberOfPlaceholders = 0; numberOfPlaceholders < 3; ++numberOfPlaceholders) {
+    for (let numberOfPlaceholders = 0; numberOfPlaceholders < 3; ++numberOfPlaceholders) {
       const exampleInput = [
         '# <STRUCTURE-BELOW-IS-TO-BE-DOCUMENTED>',
         '[args] > simple',
@@ -64,6 +64,9 @@ describe('generate_comments', () => {
       verifyGeneratedOutput(stdout, home, outputFilePath, expectedContents);
     }
     done();
+  });
+  it('does not leak placeholder counters into global scope', () => {
+    assert.strictEqual(globalThis.numberOfPlaceholders, undefined);
   });
 });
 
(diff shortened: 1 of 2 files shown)


test/commands/test_generate_comments.js
```diff
@@ -23,7 +23,7 @@ describe('generate_comments', () => {
   });
   it('fills output depending on the number of placeholders in the input code', (done) => {
     const home = makeHome();
-    for (numberOfPlaceholders = 0; numberOfPlaceholders < 3; ++numberOfPlaceholders) {
+    for (let numberOfPlaceholders = 0; numberOfPlaceholders < 3; ++numberOfPlaceholders) {
```
Thayorns, line 26: i would rather rename this variable to be less composite

Thayorns (commented):
@yegor256 this PR looks great, but unfortunately the issue is already closed.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+299](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #978, opened 2026-06-06
test: remove redundant mvnw catch

### LINKED ISSUE #970: `count` function incorrectly rethrows ENOENT errors in `test/test_mvnw.js`
The catch block inside the local `count` function in `test/test_mvnw.js:80` branches on the error code and then throws either way. Line 81 says `if (error.code === 'ENOENT') { throw error; }` and line 84 is a bare `throw error` right after it, so both paths do exactly the same thing. The conditional decides nothing, and the whole `try/catch` collapses to rethrowing every error it catches.

The test is named `should handle ENOENT race condition in count function` and its assertion claims an ENOENT is handled gracefully, yet this block rethrows ENOENT like any other error. So the conditional does not do what the test name promises; it reads as if a graceful branch was meant to live there and was lost, leaving dead code behind. The test still passes only because the problematic file is unlinked before `count` runs, so the ENOENT path is never actually taken.

The minimal fix is to drop the redundant `if` on lines 81-83 so the catch is a plain `throw error`. If the intent was really to survive the race-condition ENOENT, that first branch should skip the missing entry instead of throwing a second time.

999axel999:
## Summary
- Remove the redundant local `count` try/catch in `test/test_mvnw.js`.
- Keep the test behavior unchanged while dropping the dead ENOENT branch.

Closes #970

## Tests
- `node ..\objectionary__eoc_fix\node_modules\mocha\bin\mocha.js test\test_mvnw.js --grep "ENOENT race" --timeout 1200000`
- `node ..\objectionary__eoc_fix\node_modules\eslint\bin\eslint.js test\test_mvnw.js`
- `git diff --check`
- `git merge-tree --write-tree origin/master HEAD`

### DIFF
--- a/test/test_mvnw.js
+++ b/test/test_mvnw.js
@@ -37,17 +37,10 @@ describe('mvnw', () => {
       if (fs.existsSync(dir)) {
         for (const f of fs.readdirSync(dir)) {
           const next = path.join(dir, f);
-          try {
-            if (fs.statSync(next).isDirectory()) {
-              curr = count(next, curr);
-            } else {
-              curr++;
-            }
-          } catch (error) {
-            if (error.code === 'ENOENT') {
-              throw error;
-            }
-            throw error;
+          if (fs.statSync(next).isDirectory()) {
+            curr = count(next, curr);
+          } else {
+            curr++;
           }
         }
       }


Thayorns (commented):
@yegor256 this PR can be closed as a duplicate. The issue is closed already

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (15) hits-of-code; +4 to give you at least something.  Your running score is [+292](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #980, opened 2026-06-06
test: cover parser version request behavior

### LINKED ISSUE #971: Duplicate tests in `test/test_parser_version.js` provide false coverage for URL construction and network-error handling
In `test/test_parser_version.js` the test `constructs correct Maven Central URL` at line 63 is a verbatim copy of `returns true for a known valid version` at line 26. Both call `parserVersion.exists('0.28.11')` and assert the result is `true`; the only thing that differs is the failure message. The test never looks at the URL at all, so it cannot tell a correctly built URL from a broken one. If `exists` in `src/parser-version.js:40` assembled the wrong path, this test would pass or fail exactly like the plain happy-path test sitting two cases above it, which makes its name a promise it does not keep.

`handles network errors gracefully` at line 68 is the same happy-path call wrapped in `assert.doesNotThrow`. The describe block gates on connectivity with `before(weAreOnline)` at line 11, so the case only runs when online, `exists('0.28.11')` takes the 200 branch, and the network-error path is never reached. On top of that, `exists` already swallows every exception in its own `try/catch` and returns `false` (`src/parser-version.js:43`), so `assert.doesNotThrow` is satisfied for any argument at all. The assertion can never fail, so it verifies nothing.

The result is false coverage. Two cases named for URL construction and for network-error handling assert nothing beyond what `returns true for a known valid version` already covers, so a regression in either of those behaviors would still land green. Either delete the two duplicates, or make them real: assert against the URL that `exists` builds and drive the `catch` branch so it returns `false`, which would mean letting the HTTP call be injected instead of the hard-wired `sync-request` call.

999axel999:
## Summary
- Replace the duplicate parser-version URL test with a request-stubbed assertion for the exact Maven Central POM URL and timeout options.
- Replace the no-op network-error test with a request-stubbed failure that verifies `exists()` returns `false` from its catch branch.

Closes #971

## Tests
- `node ..\objectionary__eoc_fix\node_modules\mocha\bin\mocha.js test\test_parser_version.js --grep 'constructs correct Maven Central URL|handles network errors gracefully' --timeout 1200000` -> 2 passing
- `node ..\objectionary__eoc_fix\node_modules\eslint\bin\eslint.js test\test_parser_version.js` -> passed
- `git diff --check` -> passed
- `git merge-tree --write-tree origin/master HEAD` -> passed

### DIFF
--- a/test/test_parser_version.js
+++ b/test/test_parser_version.js
@@ -6,6 +6,41 @@
 const assert = require('assert');
 const parserVersion = require('../src/parser-version');
 const {weAreOnline} = require('./helpers');
+const parserModule = '../src/parser-version';
+const requestModule = 'sync-request';
+
+/**
+ * Load parser-version with a temporary sync-request implementation.
+ * @param {Function} request Request stub
+ * @param {Function} check Test callback
+ * @return {*} Callback result
+ */
+const withRequestStub = (request, check) => {
+  const parserPath = require.resolve(parserModule),
+    requestPath = require.resolve(requestModule),
+    originalParser = require.cache[parserPath],
+    originalRequest = require.cache[requestPath];
+  delete require.cache[parserPath];
+  require.cache[requestPath] = {
+    'id': requestPath,
+    'filename': requestPath,
+    'loaded': true,
+    'exports': request,
+  };
+  try {
+    return check(require(parserModule));
+  } finally {
+    delete require.cache[parserPath];
+    if (originalParser) {
+      require.cache[parserPath] = originalParser;
+    }
+    if (originalRequest) {
+      require.cache[requestPath] = originalRequest;
+    } else {
+      delete require.cache[requestPath];
+    }
+  }
+};
 
 describe('parser-version', () => {
   before(weAreOnline);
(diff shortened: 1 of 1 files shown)


test/test_parser_version.js
```diff
@@ -6,6 +6,41 @@
...
+const parserModule = '../src/parser-version';
+const requestModule = 'sync-request';
+
+/**
+ * Load parser-version with a temporary sync-request implementation.
+ * @param {Function} request Request stub
+ * @param {Function} check Test callback
+ * @return {*} Callback result
+ */
+const withRequestStub = (request, check) => {
```
Thayorns, line 18: I would recommend to rename this function to be less composite

test/test_parser_version.js
```diff
@@ -6,6 +6,41 @@
...
 const parserVersion = require('../src/parser-version');
 const {weAreOnline} = require('./helpers');
+const parserModule = '../src/parser-version';
+const requestModule = 'sync-request';
+
+/**
+ * Load parser-version with a temporary sync-request implementation.
+ * @param {Function} request Request stub
+ * @param {Function} check Test callback
+ * @return {*} Callback result
```
Thayorns, line 16: The doc says `@return {*}` — returning an arbitrary passthrough invites null propagation

test/test_parser_version.js
```diff
@@ -6,6 +6,41 @@
...
+ */
+const withRequestStub = (request, check) => {
+  const parserPath = require.resolve(parserModule),
+    requestPath = require.resolve(requestModule),
+    originalParser = require.cache[parserPath],
```
Thayorns, line 21: `require.cache` — this is shared global state and reflection-like introspection of the module system (require.resolve, cache surgery). Type/internals introspection is prohibited

Thayorns (commented):
A test should inject the request collaborator through a constructor, not by hijacking `require`.

Also, this PR is a duplicate, the issue is closed already

yegor256:

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (3) comments.  Your running score is [+228](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #984, opened 2026-06-08
fix(#971): make parser-version URL and network-error tests assert real behavior

### LINKED ISSUE #971: Duplicate tests in `test/test_parser_version.js` provide false coverage for URL construction and network-error handling

Thayorns:
The `exists()` suite in `test/test_parser_version.js` had two cases that promised something their assertions never checked. "constructs correct Maven Central URL" was a verbatim copy of the plain valid-version test: it called `exists('0.28.11')` and asserted the result was `true`, never once looking at the URL. "handles network errors gracefully" wrapped the same happy-path call in `assert.doesNotThrow`, but the describe block gates on connectivity, so the call always took the 200 branch — and `exists` already swallows exceptions internally, so the assertion could not fail for any input. Both were green regardless of whether the URL or the catch branch actually worked.

To make them real, `parser-version.js` now exposes a small `url(ver)` method, and `exists(ver, fetch)` accepts an injected HTTP call that defaults to `sync-request`. The URL test asserts the exact path the code builds, with no network involved. The network-error test injects a fetch that throws and asserts `exists` returns `false`, which finally drives the `catch` branch. A third case injects a non-200 response to cover that branch deterministically too.

Closes #971

### DIFF
--- a/src/parser-version.js
+++ b/src/parser-version.js
@@ -27,18 +27,26 @@ const request = require('sync-request'),
       return version.value;
     },
     /**
+     * Build the Maven Central URL of a parser version POM.
+     * @param {String} ver - Version to locate, for example '0.23.1'
+     * @return {String} Full URL of the eo-maven-plugin POM
+     */
+    url(ver) {
+      const repo = 'org/eolang/eo-maven-plugin',
+        artifactId = 'eo-maven-plugin';
+      return `https://repo.maven.apache.org/maven2/${repo}/${ver}/${artifactId}-${ver}.pom`;
+    },
+    /**
      * Check if a specific parser version exists in Maven Central.
      * @param {String} ver - Version to check, for example '0.23.1'
+     * @param {function} fetch - HTTP call, defaults to sync-request
      * @return {Boolean} True if version exists, false otherwise
      */
-    exists(ver) {
+    exists(ver, fetch = request) {
       let result;
       if (ver && ver !== 'undefined') {
         try {
-          const repo = 'org/eolang/eo-maven-plugin',
-            artifactId = 'eo-maven-plugin',
-            url = `https://repo.maven.apache.org/maven2/${repo}/${ver}/${artifactId}-${ver}.pom`,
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 ready for review

0crat:
@thayorns Thanks for the contribution! You've earned **+16 points** following our bonus policy - you received the base 16 points for providing reviews, plus additional points for your code contribution while staying within our quality thresholds. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+144](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #985, opened 2026-06-08
fix(#969): declare test helper variables so they dont leak onto globalThis

### LINKED ISSUE #969: Undeclared variables in `test/commands/java/test_pipeline.js` causing potential global scope conflict

Thayorns:
The helper variables in two test files were assigned without a declaration, so each bare assignment landed on `globalThis` instead of staying local. The three `java/pipeline` tests shared one `calls` array through the global object, and the `generate_comments` loops did the same with their counter. They passed only because every test reset the value at its top; the moment a test forgot to reset, or the file gained `'use strict'`, the behaviour would change or throw.

This declares each one in place: `const calls` and `const opts` in `test/commands/java/test_pipeline.js`, and `let placeholders` in the two loops of `test/commands/test_generate_comments.js`. The counter is also renamed from `numberOfPlaceholders` to keep it within the two-word limit. Two guards assert the globals stay clean between tests, so the leak cannot creep back.

Closes #969

### DIFF
--- a/test/commands/java/test_pipeline.js
+++ b/test/commands/java/test_pipeline.js
@@ -12,11 +12,11 @@ describe('java/pipeline', () => {
       register: {goals: () => ['eo:register']},
       assemble: {goals: () => ['eo:assemble']},
     };
-    calls = [];
+    const calls = [];
     const maven = function maven(command) {
       calls.push(command);
     }
-    opts = {sources: 'sources-dir', target: 'target-dir'};
+    const opts = {sources: 'sources-dir', target: 'target-dir'};
     await pipeline(coms, ['register', 'assemble'], opts, maven)
     assert.deepStrictEqual(
       calls[0],
@@ -45,11 +45,11 @@ describe('java/pipeline', () => {
         extras: (opts) => [`-Deo.failOnWarning=${opts.easy ? 'false' : 'true'}`],
       },
     };
-    calls = [];
+    const calls = [];
     const maven = function maven(command) {
       calls.push(command);
     }
-    opts = {
+    const opts = {
       sources: 'lint-sources-dir',
       target: 'lint-target-dir',
       newParser: true,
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 ready for review

0crat:
@thayorns Great work on your contribution! 🎉 You've earned **+16 points** and your total score has reached **[+128](https://www.eolang.org/zerocracy/objectionary-vitals.html)**. [Keep up the excellent pace](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) and don't forget to monitor your progress on your [Zerocracy account](https://www.zerocracy.com/).

PR #989, opened 2026-06-08
fix(#970): handle the ENOENT race in count instead of rethrowing it twice

### LINKED ISSUE #970: `count` function incorrectly rethrows ENOENT errors in `test/test_mvnw.js`

Thayorns:
The `count` helper in `test/test_mvnw.js` caught every error and then threw it again from both arms of the conditional: `if (error.code === 'ENOENT') { throw error; } throw error;`. The branch chose nothing, so the whole `try/catch` did no more than rethrow. The test is named after handling an ENOENT race, and it asserts the error is handled gracefully, yet the code rethrew ENOENT like anything else. It only passed because the file was unlinked before `count` ran, so the race was never reached.

This makes the helper survive the race it claims to: an entry that disappears between `readdirSync` and `statSync` is skipped, while any other error still propagates. The test now triggers a real ENOENT through a dangling symlink and checks that the count tallies the surviving class instead of throwing. A fresh temp directory replaces the manual setup and teardown.

Closes #970

### DIFF
--- a/test/test_mvnw.js
+++ b/test/test_mvnw.js
@@ -6,6 +6,7 @@
 const {mvnw, flags} = require('../src/mvnw');
 const assert = require('assert');
 const fs = require('fs');
+const os = require('os');
 const path = require('path');
 const {execSync} = require('child_process');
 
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 ready for review

0crat:
@thayorns Hey there! 🎉 Nice work on that contribution - you just scored **+16 points**! Your code hit the sweet spot with good size and solid reviews, which is exactly what our [bonus system](https://www.yegor256.com/2014/11/14/hits-of-code.html) rewards. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) these quality contributions coming and you'll keep climbing - you're already at [+176](https://www.eolang.org/zerocracy/objectionary-vitals.html) total! Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for the latest updates.

PR #995, opened 2026-06-14
Add trailing newline to actionlint workflow

Thayorns:
The yamllint check (`ibiqlik/action-yamllint`) has been failing on every pull request, not just changes that touch the workflow. The cause is `.github/workflows/actionlint.yml`, whose final line `- uses: reviewdog/action-actionlint@v1.72.0` had no terminating newline after commit b3af63c. yamllint reports it as `new-line-at-end-of-file` and since the job lints the whole repository, the red check spreads to unrelated work.

This adds the single trailing newline so the file ends cleanly and the check goes green again.

The change is deliberately one line. A missing trailing newline cannot be meaningfully unit-tested, so there is nothing to add beyond the fix itself.


Thayorns:
@yegor256 take a look

PR #1011, opened 2026-06-23
Split fmt test into independent per-case tests

### LINKED ISSUE #992: `fmt` test inefficiently runs four Maven builds serially in single `it()` in `test/commands/test_fmt.js`
The test "formats EO files according to expected patterns" at `test/commands/test_fmt.js:133` loops over four `testCases` and calls `fmt()` for each, and every `fmt()` spawns a fresh `node src/eoc.js fmt` that boots a cold Maven JVM and runs the whole format pipeline. Four full builds end up serialized inside a single test, which is why this one case takes around 47 seconds on its own, the slowest in the suite.

Because the four builds live in one `it()`, Mocha cannot spread them across workers under `--parallel`, so the suite has no way to overlap them. The test also reuses one `temp/test-fmt/simple` directory for all four cases and carries eight assertions in the loop; when an early case fails the rest never run, and the hand-written `Test case N failed` message stands in for the per-case name that a real test would have.

Thayorns:
The `fmt` formatting test packed four separate format scenarios into a single `it()` that looped over the cases. Each iteration spawns a cold Maven JVM and runs the whole pipeline, so four full builds ran serialized inside one test, reusing the same `temp/test-fmt/simple` directory and carrying eight assertions. When an early case failed the rest never ran, and a hand-written `Test case N failed` message stood in for a real per-case name.

This splits the data table into one named `it()` per case, each with its own isolated temp directory. Failures no longer cascade, every scenario reports under its own sentence, and the cases are independent units that a parallel runner can schedule on their own.

I also added an idempotency edge case: formatting an already-formatted file a second time must leave it unchanged.

Closes #992

### DIFF
--- a/test/commands/test_fmt.js
+++ b/test/commands/test_fmt.js
@@ -10,10 +10,11 @@ const { runSync, parserVersion, homeTag, weAreOnline } = require('../helpers')
 
 /**
  * Prepare test directory and return the home path
+ * @param {string} name - Unique sub-directory name for this case
  * @return {Array} Array containing source and target directory paths
  */
-function prepareTestDirectory () {
-  const home = path.resolve('temp/test-fmt/simple')
+function prepareTestDirectory (name) {
+  const home = path.resolve('temp/test-fmt', name)
   const source = path.resolve(home, 'src')
   const target = path.resolve(home, 'target')
   fs.rmSync(home, { recursive: true, force: true })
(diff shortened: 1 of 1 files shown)


Thayorns:
@yegor256 review please

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+266](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1019, opened 2026-06-27
Reject in mvnw on non-zero Maven exit instead of killing the process

### LINKED ISSUE #999: `eoc lint` fails to provide `--easy` hint on linting errors
**What happens:** When `eoc lint` detects errors, the process exits via `process.exit(1)` inside `mvnw()` (see `src/mvnw.js`, line 81). The `reject` callback of the returned Promise is never called, so the `catch` block in `src/commands/lint.js` — which is supposed to rethrow with the message "There are errors and/or warnings; you may disable warnings via the --easy option" — is dead code that can never execute. The user only sees Maven's raw exit-code message and never learns about the `--easy` flag.

**What should happen:** When Maven exits with a non-zero code, `mvnw()` should call `reject(new Error(...))` instead of (or before) `process.exit(1)`, so that command-level `catch` blocks can add user-friendly guidance such as the `--easy` hint before the process terminates.

Thayorns:
When `eoc lint` found problems, `mvnw()` terminated the whole process with `process.exit(1)` the moment Maven returned a non-zero code. The returned promise was never rejected, so the `catch` blocks in the lint command never ran. Users only ever saw Maven's raw exit-code line and were never told about the `--easy` flag.

Now `mvnw()` rejects with an `Error` carrying the exit-code context, so command-level `catch` blocks can add their guidance (the `--easy` hint) before the process ends. In the quiet path the progress spinner is stopped first so it does not keep printing after a failure.

Since the command actions are async, `eoc.js` switches to `parseAsync`. We deliberately rely on the native Node.js handling here rather than a manual `catch`: an unhandled rejection from `parseAsync` is escalated by Node's default `--unhandled-rejections=throw` mode, which prints the error message and exits with a non-zero code. This keeps the entry point a clean one-liner while still surfacing the error to the user.

Four tests cover the new behaviour: quiet and verbose runs both reject instead of tearing down the test process, the rejection is a real `Error` that callers can wrap as a cause, and `mvnw` can run again after a previous run failed.

Closes #999

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -378,13 +378,7 @@ program.command('fmt')
   });
 
 if (require.main === module) {
-  try {
-    program.parse(process.argv);
-  } catch (e) {
-    console.error(e.message);
-    console.debug(e.stack);
-    process.exit(1);
-  }
+  program.parseAsync(process.argv);
 }
 
 module.exports.commandsDescription = function commandsDescription() {
--- a/src/mvnw.js
+++ b/src/mvnw.js
@@ -110,20 +110,20 @@ module.exports.mvnw = function(args, tgt, batch) {
         start();
       }
       result.on('close', (code) => {
-        if (code !== 0) {
-          console.error(`The command "${cmd}" exited with #${code} code`);
-          process.exit(1);
-        }
         if (!batch) {
           stop();
+        }
+        if (code !== 0) {
+          reject(new Error(`The command "${cmd}" exited with #${code} code`));
+          return;
         }
         resolve(args);
       });
     } else {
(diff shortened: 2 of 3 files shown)


Thayorns:
@yegor256 review please

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+334](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1020, opened 2026-06-27
Skip parser version check on network errors instead of a false-negative absence

### LINKED ISSUE #998: `eoc` outputs incorrect "not available in Maven Central" error during network unavailability
**What happens:** `src/parser-version.js` `exists()` catches network errors silently and returns `false`. Then `src/mvnw.js` `flags()` checks `!parserVersion.exists(opts.parser)` and, on a `false` result, prints "Parser version X is not available in Maven Central" and calls `process.exit(1)`. This fires even when the parser version is perfectly valid — e.g., the bundled default in `eo-version.txt` — but Maven Central is temporarily unreachable or the user is offline. The error message is factually wrong: the version exists, the network does not.

**What should happen:** `exists()` should distinguish between a confirmed 404 (version truly absent) and a network/timeout error. On a network error the check should be skipped with a warning ("cannot verify parser version, proceeding anyway") rather than aborting with a false-negative error message.

Thayorns:
`parser-version.js`'s `exists()` caught every error from the HTTP call and returned `false`, lumping a real "version absent" answer together with "could not reach Maven Central". `mvnw.js` then treated that `false` as proof of absence and aborted with "Parser version X is not available in Maven Central" plus `process.exit(1)` — even for the bundled default version, when the only real problem was that the user was offline or Maven Central was briefly unreachable.

`sync-request` throws only on connection or timeout failures; an HTTP 404 comes back as a normal response with a status code. That gives a clean line to draw: a thrown error means the network failed, so `exists()` now warns ("cannot verify parser version, proceeding anyway") and returns `true` so the run continues. A non-200 response is still a confirmed absence and returns `false`, so the genuine "not available" error keeps working. No change is needed in `mvnw.js`.

Tests cover both halves of the split: network error and timeout both skip the check, a 200 confirms the version, a 404 reports it absent, plus the POM URL it queries and that an undefined version never touches the network.

Closes #998

### DIFF
--- a/src/parser-version.js
+++ b/src/parser-version.js
@@ -4,6 +4,7 @@
  */
 
 const {XMLParser} = require('fast-xml-parser');
+const colors = require('colors');
 const request = require('sync-request'),
 
   /**
@@ -40,7 +41,7 @@ const request = require('sync-request'),
      * Check if a specific parser version exists in Maven Central.
      * @param {String} ver - Version to check, for example '0.23.1'
      * @param {function} fetch - HTTP call, defaults to sync-request
-     * @return {Boolean} True if version exists, false otherwise
+     * @return {Boolean} True if version exists or cannot be verified, false if confirmed absent
      */
     exists(ver, fetch = request) {
       let result;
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 review please

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+350](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1021, opened 2026-06-27
Make --language option case-insensitive and accept aliases (js/java)

Mirletti:
## Problem

The global `--language` option is matched against the exact canonical
platform names `Java` and `JavaScript` only. The natural, lower-case
inputs that users are most likely to type fail with a confusing error:

```
$ eoc --language=js transpile
Unknown platform js
$ eoc --language=java compile
Unknown platform java
```

Only `--language=Java` / `--language=JavaScript` (exact case) work today.

## Fix

A small `platform()` resolver maps the user-provided value to a canonical
platform key. The lookup is case-insensitive and accepts the common
aliases `java`, `js` and `javascript`, so all of these now work:

```
--language=js  --language=JS  --language=javascript  --language=JavaScript
--language=java  --language=JAVA  --language=Java
```

Unknown names are returned unchanged, so an invalid value is still
reported clearly (e.g. `Unknown platform Eiffel`) and the existing
behaviour/tests for that case are preserved. The `--language` help text
now also documents the accepted values.

## Tests

* Unit tests for the exported `platform()` resolver (canonical names,
  lower-case aliases, mixed case, and pass-through of unknown names).
* CLI integration tests driving the `clean` command with `--language=js`
  and `--language=JAVA`, plus an unknown-language failure assertion.

`npx eslint` passes clean.

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -63,6 +63,20 @@ const {program} = require('commander'),
     [language.js]: require('./commands/js/pipeline'),
   };
 
+/**
+ * Every language name a user may type, mapped (in lower case) to its
+ * canonical platform key. Derived from `language` so the accepted names
+ * and the platforms can never drift apart. For each `[alias, canonical]`
+ * pair we accept the short alias (e.g. `js`) and the canonical name
+ * lower-cased (e.g. `javascript`); `select()` lower-cases the input, so
+ * mixed-case spellings such as `JavaScript` are matched too.
+ */
+const platforms = {};
+for (const [alias, canonical] of Object.entries(language)) {
+  platforms[alias] = canonical;
+  platforms[canonical.toLowerCase()] = canonical;
+}
+
 if (process.argv.includes('--verbose')) {
   tinted.enable('debug');
   console.debug('Debug output is turned ON');
@@ -103,7 +117,7 @@ program
   .option('--parser <version>', 'Set the version of EO parser to use', parser)
   .option('--latest', 'Use the latest parser version from Maven Central')
   .option('--alone', 'Just run a single command without dependencies')
-  .option('-l, --language <name>', 'Language of target execution platform', language.java)
+  .option('-l, --language <name>', 'Language of target execution platform (Java or JavaScript, case-insensitive)', language.java)
   .option('-b, --batch', 'Run in batch mode, suppress interactive messages')
   .option('--no-color', 'Disable colorization of console messages')
   .option('--track-transformation-steps', 'Save intermediate XMIR files')
@@ -418,12 +432,7 @@ function pin(opts) {
  * @return {Object} - commands
  */
 function coms() {
-  const lang = program.opts().language,
-    hash = commands[lang];
-  if (hash === undefined) {
-    throw new Error(`Unknown platform ${lang}`);
-  }
-  return hash;
+  return select(commands, program.opts().language);
 }
 
 /**
@@ -431,10 +440,24 @@ function coms() {
  * @return {Function} - pipeline function
  */
 function pipe() {
-  const lang = program.opts().language;
-  const pipeline = pipelines[lang];
-  if (pipeline === undefined) {
+  return select(pipelines, program.opts().language);
+}
+
+/**
+ * Resolve the entry registered for the requested language, matching the
+ * name case-insensitively and accepting aliases (e.g. `js`, `java`). The
+ * lookup and its validation live here, so callers never repeat the
+ * "unknown platform" check.
+ *
+ * @param {Object} registry - Map keyed by canonical platform name
+ * @param {String} lang - Language name as provided on the command line
+ * @return {*} The entry registered for the resolved platform
+ * @throws {Error} If the language does not resolve to a known platform
+ */
+function select(registry, lang) {
+  const found = registry[platforms[String(lang).toLowerCase()]];
+  if (found === undefined) {
     throw new Error(`Unknown platform ${lang}`);
   }
-  return pipeline;
+  return found;
 }
--- a/test/test_eoc.js
+++ b/test/test_eoc.js
@@ -40,6 +40,30 @@ describe('eoc', () => {
 });
 
 describe('eoc', () => {
+  const {spawnSync} = require('child_process'),
+    path = require('path'),
+    eoc = function(...args) {
+      return spawnSync('node', [path.resolve('./src/eoc.js'), '--batch', ...args]);
+    };
+  it('accepts the "js" alias for the --language option', (done) => {
+    assert.strictEqual(eoc('--language=js', 'clean').status, 0);
+    done();
+  });
+  it('accepts the full "javascript" name for the --language option', (done) => {
+    assert.strictEqual(eoc('--language=javascript', 'clean').status, 0);
+    done();
+  });
+  it('accepts a mixed-case value for the --language option', (done) => {
+    assert.strictEqual(eoc('--language=JAVA', 'clean').status, 0);
+    done();
+  });
+  it('rejects an unknown --language value', (done) => {
+    assert.notStrictEqual(eoc('--language=Eiffel', 'clean').status, 0);
+    done();
+  });
+});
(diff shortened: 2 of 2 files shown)


0crat:
@mirletti It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`case-insensitive-language`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`1020`". Your running score is [-24](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

src/eoc.js
```diff
@@ -432,9 +432,32 @@ function coms() {
...
+ * so the caller can still report them as "Unknown platform".
+ *
+ * @param {String} lang - Language name as provided on the command line
+ * @return {String} Canonical platform key, e.g. 'Java' or 'JavaScript'
+ */
+function platform(lang) {
+  const aliases = {
+    'java': language.java,
+    'js': language.js,
+    'javascript': language.js,
```
Thayorns: This map duplicates the existing `language` object: its keys (`java`, `js`) are already `language`'s keys and the values are `language`'s values — only `javascript` is genuinely new. The knowledge now lives in two places, so adding a platform later means editing `language`, `commands`, `pipelines` *and* this map, and the duplicate will drift out of sync silently.

src/eoc.js
```diff
@@ -432,9 +432,32 @@ function coms() {
...
+  };
+  const key = String(lang).toLowerCase();
+  return Object.hasOwn(aliases, key) ? aliases[key] : lang;
```
Thayorns: For unknown input this returns the value unchanged, so `platform('Eiffel') === 'Eiffel'`, yet the `@return` just above promises a "Canonical platform key, e.g. 'Java' or 'JavaScript'". Resolution and validation end up split between this function and its callers.

src/eoc.js
```diff
@@ -432,9 +432,32 @@ function coms() {
...
+}
+
+module.exports.platform = platform;
```
Thayorns: This exposes an internal helper purely so the unit test can reach it, which goes against avoiding non-interface public methods.

src/eoc.js
```diff
@@ -432,9 +432,32 @@ function coms() {
  */
 function pipe() {
   const lang = program.opts().language;
-  const pipeline = pipelines[lang];
+  const pipeline = pipelines[platform(lang)];
```
Thayorns: This `platform(lang)` + `undefined → throw 'Unknown platform'` logic is repeated verbatim in `coms()` (line 422).

Thayorns (changes requested):
The feature itself — case-insensitive `--language` with `js`/`java` aliases — is reasonable. My concern is the alias resolution, which re-declares state that already lives in the `language` object. See the inline notes on `src/eoc.js`.

Briefly on the tests: several cases carry multiple assertions (and use loose `assert.equal` rather than `strictEqual`), and the CLI cases assert on stdout/stderr text (`'no need to delete it'`, `'Unknown platform Eiffel'`), which runs against the project's convention of not testing logs.

Mirletti:
Thanks for the review @Thayorns — all points addressed in the latest push (history squashed to a single clean commit):

- **Alias map duplicated `language`.** Removed the hand-written map. The lookup is now a `platforms` table *derived* from the existing `language` object, so aliases and platforms can't drift: for every `[alias, canonical]` entry both `alias.toLowerCase()` and `canonical.toLowerCase()` map to `canonical`. Adding a platform now means editing only `language` (plus `commands`/`pipelines`, as before) — the lookup updates automatically.
- **`platform()` returned unknown input unchanged / split resolution & validation.** That function is gone. Resolution and the "unknown platform" validation now live together in a single `select(registry, lang)` helper that returns the resolved entry or throws.
- **Exported internal helper just for tests.** Removed the `module.exports.platform` export — there is no non-interface public method anymore.
- **`platform(lang) + undefined → throw` duplicated in `coms()` and `pipe()`.** De-duplicated: both now delegate to `select()`, which owns the check.
- **Tests.** Dropped the unit tests that reached into the internal helper. The remaining CLI tests now assert a single thing each — the process exit status (`strictEqual`/`notStrictEqual`) — rather than stdout/stderr text, matching the convention of not testing logs.

src/eoc.js
```diff
@@ -61,7 +61,21 @@ const {program} = require('commander'),
...
+ * platforms never drift apart: both the short alias (e.g. `js`) and the
+ * full canonical name (e.g. `JavaScript`) resolve to the same key used
+ * in `commands` and `pipelines`.
+ */
+  platforms = Object.fromEntries(
+    Object.entries(language).flatMap(([alias, canonical]) => [
+      [alias.toLowerCase(), canonical],
+      [canonical.toLowerCase(), canonical],
+    ])
+  );
```
Thayorns: Two things here, both about keeping this readable:

1. **`alias.toLowerCase()` is redundant.** `alias` is a key of `language` (`java`, `js`), which is already authored in lowercase, so this call is a no-op today and silently couples the map to "keys are always lowercase." It also makes the `java` case emit the identical pair `['java', 'Java']` twice (since `'java'.toLowerCase() === 'Java'.toLowerCase()`) — the duplicate only survives because the object dedupes keys. `[alias, canonical]` is equally correct and clearer. (`canonical.toLowerCase()` and the `toLowerCase()` in `select()` are doing real work — keep those.)

2. **The `Object.fromEntries(Object.entries(...).flatMap(...))` chain packs a lot into one expression.** The derivation idea is good, but three nested transforms plus the redundant lower-casing make the intent hard to follow at a glance. Removing the redundant `alias.toLowerCase()` already helps; consider whether a small explicit builder (or a short comment naming the three accepted spellings — short alias, canonical, canonical-lowercased) would read more plainly than the one-liner.

Thayorns (changes requested):
Worth saying the direction is right: deriving the lookup from `language` via `Object.fromEntries` removes the hand-maintained `aliases` map from the earlier revision, so aliases and platforms can no longer drift — that resolves my main concern from the first round. `select()` consolidating the "unknown platform" guard and dropping the test-only `platform` export are good fixes too.

Mirletti:
Thanks @Thayorns — both readability points addressed:

- **Dropped the redundant `alias.toLowerCase()`.** The keys of `language` are already authored in lower case, so it was a no-op that also emitted the `['java', 'Java']` pair twice. The builder now uses `alias` directly.
- **Unpacked the one-liner into a plain builder.** Replaced the `Object.fromEntries(Object.entries(...).flatMap(...))` chain with a small explicit loop, plus a comment naming the accepted spellings (short alias, canonical lower-cased; `select()` lower-cases the input, so mixed-case names match too):

```js
const platforms = {};
for (const [alias, canonical] of Object.entries(language)) {
  platforms[alias] = canonical;
  platforms[canonical.toLowerCase()] = canonical;
}
```

`canonical.toLowerCase()` and the `toLowerCase()` in `select()` are kept, as you noted they do real work. Behaviour is unchanged (`js`/`java`/`javascript`/`JavaScript`/`JAVA` all resolve; unknown values still fail with `Unknown platform <name>`).

Thayorns (approved):
Looks perfect 👍

0crat:
@mirletti Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+22](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+237](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1022, opened 2026-06-27
Fix literal "%s" in docs summary log message

Mirletti:
## Problem

The `docs` command logs where the summary XML was written using a
printf-style call:

```js
tracked.print('Summary XML generated at %s', path.relative(process.cwd(), summary));
```

However, the `print` helper returned by `elapsed()` takes a single
`message` argument and performs no `%s` substitution — it just appends
the elapsed time. As a result the `%s` is printed literally and the
path is lost:

```
Summary XML generated at %s in 95ms
```

This is the only `tracked.print` call in the codebase that uses
printf-style formatting; every other call already uses a template
literal.

## Fix

Use a template literal, matching the rest of the codebase, so the real
path is shown:

```
Summary XML generated at .eoc/docs/summary.xml in 95ms
```

## Tests

Added a test that drives the `docs` command directly, captures the log
output, and asserts the summary line contains the actual path and no
literal `%s`. The test fails against the old code and passes with the
fix.

`npx eslint` passes clean.

### DIFF
--- a/src/commands/docs.js
+++ b/src/commands/docs.js
@@ -119,6 +119,7 @@ function wrapHtml(name, html, css) {
 /**
  * Command to generate documentation.
  * @param {Hash} opts - All options
+ * @return {Promise<String>} Resolves to the message reporting the summary path
  */
 module.exports = function(opts) {
   return elapsed(async (tracked) => {
@@ -173,8 +174,9 @@ module.exports = function(opts) {
       }
       lines.push('</eodoc>');
       fs.writeFileSync(summary, lines.join('\n'));
-      tracked.print('Summary XML generated at %s', path.relative(process.cwd(), summary));
+      const located = tracked.print(`Summary XML generated at ${path.relative(process.cwd(), summary)}`);
       tracked.print(`Documentation generation completed in the ${output} directory`);
+      return located;
     } catch (error) {
       console.error('Error generating documentation:', error);
       throw error;
--- a/test/commands/test_docs.js
+++ b/test/commands/test_docs.js
@@ -7,6 +7,7 @@ const assert = require('assert');
 const fs = require('fs');
 const path = require('path');
 const {runSync} = require('../helpers');
+const generateDocs = require('../../src/commands/docs');
 
 describe('docs', () => {
   const home = path.resolve('temp/test-docs');
@@ -70,6 +71,22 @@ describe('docs', () => {
     assert(content.includes('<object name="test1"/>'), 'Expected object test1 in summary.xml');
     assert(content.includes('<object name="test2"/>'), 'Expected object test2 in summary.xml');
     done();
+  });
+  /**
+   * Tests that the 'docs' command reports the real summary path, rather
+   * than the literal "%s" left over by a printf-style call.
+   * @return {Promise} of the docs command, resolved to its summary message
+   */
+  it('reports the path of the generated summary', async () => {
+    const sample = path.join(parsed, 'foo', 'bar');
+    fs.mkdirSync(sample, {recursive: true});
+    fs.writeFileSync(path.join(sample, 'test1.xmir'), '<program name="test" />');
+    const expected = path.relative(process.cwd(), path.join(docs, 'summary.xml'));
+    const message = await generateDocs({target: home, sources: path.resolve(home, 'src')});
+    assert(
+      message.includes(expected),
+      `Expected "${message}" to contain "${expected}"`
+    );
   });
   /**
    * Tests that the 'docs' command generates expected comments from XMIR to HTML.


0crat:
@mirletti It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`fix-docs-summary-print`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`1021`". Your running score is [-18](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

test/commands/test_docs.js
```diff
@@ -71,6 +71,34 @@ describe('docs', () => {
     assert(content.includes('<object name="test2"/>'), 'Expected object test2 in summary.xml');
     done();
   });
+  /**
+   * Tests that the 'docs' command reports the summary path in its log,
+   * substituting the real path instead of printing a literal "%s".
+   */
+  it('prints the summary path without a literal %s placeholder', async () => {
```
Thayorns: The fix is right: `elapsed()`'s `print` (src/elapsed.js:24) takes a single `message` and does no `%s` substitution, so the printf-style call could only ever emit a literal `%s`. The template literal matches every other `tracked.print` call in the codebase.
My concern is entirely with the test, which runs against several of this project's testing conventions:
 - **It tests logs.** The test swaps out `console.info`, collects the emitted strings, and asserts on them. The bylaws state "We should not test logs."
 - **It stubs a global and cleans up in `finally`.** That's the kind of global setup/teardown and I/O stubbing the guidelines steer away from; it also mutates shared global state, which is fragile.
 - **Three assertions in one test.** The convention is one assertion, as the last statement.
 - **It asserts on the message text / the `%s` placeholder**, i.e. testing a log message's format.
 - Minor: `require('../../src/commands/docs')` inline rather than a top-level import.

 Worth noting `test/test_elapsed.js` already exercises these timing messages by asserting on the **return value** of `tracked.print(...)`, with no console interception — a precedent that avoids all of the above.

Could you rework the test along lines that fit these conventions? The source change is good to go as-is.

Thayorns (changes requested):

yegor256:
@Mirletti please, address the comment above

Mirletti:
Thanks @Thayorns — reworked the test to follow the conventions you pointed at (and the `test/test_elapsed.js` precedent):

- **No log testing / no console stubbing.** The `console.info` swap is gone. The `docs` command now *returns* its summary message (the value of `tracked.print(...)`), and the test asserts on that return value — exactly how `test/test_elapsed.js` exercises these timing messages. No globals are mutated, so there's no `try/finally` teardown either.
- **One assertion, as the last statement.** The test does its filesystem setup, then a single `assert(...)` on the returned message.
- **Top-level import.** `const generateDocs = require('../../src/commands/docs')` is now at the top of the file instead of an inline `require`.

On the source: the only addition beyond the original one-line fix is `return located;`, which gives the command a meaningful return value (the summary location) and is what makes the return-value assertion possible — mirroring how `print` itself returns its message for `test_elapsed.js` to assert on. The user-facing log order is unchanged.

The test still guards the original regression: against the old printf-style call it fails with `Expected "Summary XML generated at %s in 93ms" to contain "…/summary.xml"`, and passes with the fix.

src/commands/docs.js
```diff
@@ -119,6 +119,7 @@ function wrapHtml(name, html, css) {
 /**
  * Command to generate documentation.
  * @param {Hash} opts - All options
+ * @return {Promise.<String>} Resolves to the message reporting the summary path
```
Thayorns: The dotted `Promise.<String>` form is valid JSDoc, but it doesn't match how the rest of the codebase annotates Promise returns. The only other command that spells out the payload — `src/eo2jsw.js` — uses the un-dotted generic form `{Promise<Array.<String>>}`, and every sibling command (`assemble`, `lint`, `parse`, `print`, `audit`, `latex`, `normalize`, …) just writes bare `@return {Promise} of … task`. Please align this to the same style — either `{Promise<String>}` or the bare `{Promise}` the neighbouring commands use — so the annotation reads consistently across `src/commands`.

Thayorns (changes requested):
Requesting one change on code-style consistency. The fix itself is right — switching to a template literal so the real path is shown, and threading the message back as the return value so the test can assert on it rather than capturing logs, is the clean approach. Just one JSDoc note inline on `src/commands/docs.js`.

Mirletti:
Done — changed the annotation to the un-dotted `@return {Promise<String>}`, matching how `src/eo2jsw.js` spells out a Promise payload (`Promise<Array.<String>>`). Thanks for the review!

Thayorns (approved):
@Mirletti great job 👍!

0crat:
@mirletti Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+34](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (21) hits-of-code; +1 to give you at least something.  Your running score is [+230](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1039, opened 2026-07-09
Enforce JSDoc block alignment in ESLint

yegor256:
Adds `eslint-plugin-jsdoc` and turns on its `check-alignment` rule, so misaligned asterisks inside `/** */` blocks are now caught. The core `indent` rule only checks code tokens and skips the interior of comments, which is why the skewed blocks in `eoc.js` and a few other files slipped through. Enabling the rule flagged them, and they are realigned here.

While at it, `eoc.js` had `language`, `common`, `commands` and `pipelines` chained onto a single `const` via commas, which forced them to sit as indented continuation declarators. They are now split into separate top-level `const` statements so each global starts at column 0.

One unrelated commit rides along: `scripts/readme_automation.js` had a shebang but no executable bit, which tripped a pre-commit hook and blocked the first commit; marking it executable clears that.


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (5868 >= 100); -8 for way too many hits-of-code (5868 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1407](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1047, opened 2026-07-10
Add global --dir option to set the working directory

### LINKED ISSUE #1046: `eoc` CLI doesn't provide a global option to set the working directory
The `eoc` CLI has no way to run a build against a project that lives in another directory without first `cd`-ing into it. The global options are declared in `src/eoc.js` (lines 113-130) and every command resolves its paths from the current process working directory — for example `src/commands/register.js:17` does `path.resolve(opts.target, 'eo-foreign.json')`, and `--sources` (default `.`) and `--target` (default `.eoc`) are likewise relative to `process.cwd()`.

This is a gap because the common convention for build tools is a single flag that relocates the whole build, like `git -C <dir>` and `make -C <dir>`. Today a user has to either `cd` into the project or juggle both `--sources` and `--target` to point elsewhere, which is error-prone. (This is distinct from the closed #293, which added a per-command file-location option for the now-removed `phi`/`unphi` commands; the request here is one global working-directory flag for all commands.)

Suggested fix: add a global `-C, --dir <path>` option to the option block in `src/eoc.js` (around line 113) and, before any command action runs, call `process.chdir(opts.dir)` when the option is set, so that `--sources` and `--target` keep resolving relative to the chosen directory. Default to `.` and document it in the CLI help.

yegor256:
This adds a global `-C, --dir <path>` option to the `eoc` CLI. A `preAction` hook changes the process working directory into the given path before any command runs, so `--sources`, `--target`, and everything else keep resolving relative to it. It defaults to the current directory and exits with a clear error when the path does not exist.

Until now the only way to build a project in another directory was to `cd` into it first, or to juggle both `--sources` and `--target`. This brings `eoc` in line with the `git -C <dir>` and `make -C <dir>` convention, so `eoc --dir /path/to/project compile` just works.

Closes #1046.

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -111,6 +111,7 @@ program
   .configureHelp({sortOptions: true, sortSubcommands: true});
 
 program
+  .option('-C, --dir <path>', 'Directory where the work should be done', '.')
   .option('-s, --sources <path>', 'Directory with .EO sources', '.')
   .option('-t, --target <path>', 'Directory with all generated files', '.eoc')
   .option('--easy', 'Ignore "warnings" and only fail if there are "errors" or "critical" errors')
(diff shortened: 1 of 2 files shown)


yegor256:
@Thayorns mind taking a look when you get a chance? Happy to clarify anything about the change if something is unclear.

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1455](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1054, opened 2026-07-14
#1053: run eoc from the nix store, no home copy, jdk on wrapper path

### LINKED ISSUE #1053: `flake.nix` fails to reflect source-only changes and lacks JDK for pipeline execution
## What happens

`flake.nix` (added in #743 for #741) works for `nix build` / `nix run`, but its current implementation has several defects that will bite users as the package evolves:

1. **Stale-copy risk.** The wrapper copies the whole package out of the read-only Nix store into a mutable `$HOME/.eoc/lib` and invalidates that copy by a hash of `package-lock.json` **only**. Any change to `eoc` sources that doesn't touch the lockfile (a bugfix in `src/`, running from a patched checkout) leaves users silently executing the previous copy.

2. **Silently fragile engine patch.** `substituteInPlace package.json --replace-quiet "\"node\": \"^25.0.0\","` matches one exact byte string. When the `engines` entry is bumped (`^26.0.0`) or reformatted, `--replace-quiet` suppresses the mismatch error and the engine check quietly comes back — failing the build or the runtime depending on the nixpkgs `nodejs` version of the day.

3. **Race on first run.** Two concurrent `eoc` invocations both see a missing/old `.hash` and race `rm -rf` + `cp -r` over the same `$HOME/.eoc/lib`, corrupting the copy that one of them is executing from.

4. **No JDK on PATH.** `eoc` shells out to Java during `compile`/`link`, but the wrapper only provides `nodejs`. On a machine without a system JDK, `nix run github:objectionary/eoc` succeeds and then fails mid-pipeline.

5. **Dead code and rot risk.** The top-level `meta = ...` attribute next to `packages.default` is not a standard flake output and is ignored by Nix, and no CI job runs `nix build`, so any of the above can break unnoticed.

## What should happen

The mutable-copy design should be reconsidered, not patched. The copy exists to give the package a writable directory, but it trades away exactly what Nix is for: reproducibility (what runs depends on `$HOME` state, not on the store path), atomic upgrades and rollbacks (a half-finished `cp -r` is executable), and multi-user correctness. The right shape is the standard one for interpreted tools in nixpkgs: run straight from the read-only store, and redirect the few paths that genuinely need writing (caches, downloaded JARs) to `$XDG_CACHE_HOME`/`$HOME/.eo` — which is where per-user state belongs anyway. Concretely:

- drop the `$HOME/.eoc/lib` copy and run `node ${eolangBase}/lib/src/eoc.js` directly; if some phase still insists on writing next to the sources, that's a bug in `eoc` worth fixing rather than working around in the installer;
- remove the `engines` key format-independently (`jq 'del(.engines)'`) so the patch can't silently rot;
- wrap the binary with `makeWrapper --prefix PATH : ${pkgs.jdk}/bin` so the flake is self-contained — the whole point of installing through Nix;
- delete the stray `meta` output and add a minimal `nix build` job to CI so the flake can't break unnoticed.

I will submit a PR with the above.

morphqdd:
Closes #1053.

The flake now installs `eoc` the way Nix packages are supposed to work: the tool runs straight from the read-only store, there is no copy into `$HOME/.eoc`, and the wrapper is self contained.

What changed:

- The `$HOME/.eoc/lib` copy-out and its lockfile-hash invalidation are gone. The wrapper is a `makeWrapper` script that runs `node $out/lib/src/eoc.js` directly from the store. No stale copies, no `rm -rf` races, upgrades and rollbacks are atomic store path switches.
- The `engines` field is removed from `package.json` with `jq 'del(.engines)'` instead of a byte-exact `substituteInPlace` string, so the patch cannot silently rot when the version in `engines` is bumped or the file is reformatted.
- `jdk` is added to the wrapper PATH. Before this, `nix run` on a machine without a system Java died with `The JAVA_HOME environment variable is not defined correctly` as soon as `eoc` reached mvnw.
- The stray top-level `meta` output (ignored by Nix) is removed; the package `meta` now carries `mainProgram = "eoc"` so `nix run` resolves the binary by convention.

Verified on a NixOS machine that has no system Java at all:

- `nix build` succeeds, `eoc --version` works.
- `eoc --easy dataize app` on a hello-world program runs the whole pipeline (parse, assemble, lint, transpile) with `mvnw` executing from `/nix/store/...`, writing only into the working directory; `$HOME/.eoc` is never created.
- Control run with the previous flake on the same machine: it fails immediately with the `JAVA_HOME` error above and leaves a copy in `$HOME/.eoc/lib`.

The remaining failure in that smoke test is an upstream toolchain issue (the default `eo-maven-plugin` 0.61.0 with tag 0.59.7 fails linting on the downloaded runtime sources) and happens identically with any installation method; it is unrelated to the packaging.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,29 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+name: nix
+'on':
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+concurrency:
+  group: nix-${{ github.ref }}
+  cancel-in-progress: true
+jobs:
+  nix:
+    timeout-minutes: 30
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+      - uses: cachix/install-nix-action@v31
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+      - name: Build the flake
+        run: nix build --print-build-logs .#
+      - name: Run eoc from the store
+        run: nix run .# -- --version
--- a/flake.nix
+++ b/flake.nix
@@ -12,28 +12,17 @@
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = import nixpkgs { inherit system; };
-        srcPatched = pkgs.runCommand "patched-src" {} ''
-          cp -r ${./.} $out
-          substituteInPlace $out/package.json \
-            --replace-quiet "\"node\": \"^25.0.0\"," ""
-        '';
-        lockHash = builtins.hashFile "sha512" ./package-lock.json;
         packageJson = builtins.fromJSON (builtins.readFile ./package.json);
-        eolangBase = pkgs.buildNpmPackage {
+        srcPatched = pkgs.runCommand "eoc-src" {} ''
+          cp -r ${./.} $out
+          chmod -R u+w $out
+          ${pkgs.jq}/bin/jq 'del(.engines)' ${./.}/package.json > $out/package.json
+        '';
+      in {
+        packages.default = pkgs.buildNpmPackage {
           pname = packageJson.name;
           version = packageJson.version;
           src = srcPatched;
-
-          buildInputs = [ pkgs.nodejs ];
-          dontNpmBuild = true;
-
-          installPhase = ''
-            mkdir -p $out/lib
-            cp -r . $out/lib
-            mkdir -p $out/bin
-            ln -s $out/lib/src/eoc.js $out/bin/eoc
-            chmod +x $out/bin/eoc
-          '';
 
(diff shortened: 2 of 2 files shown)


morphqdd:
@yegor256, please take a look when you have a moment. All 16 checks are green.

yegor256:
@morphqdd how do we know all this works? maybe you can add `nix.yml` CI job for it?

morphqdd:
Added the `nix` workflow (the fifth point from #1053: a CI guard so the flake can't rot silently). It runs on push and PR, builds the flake with `nix build .#` and then runs `nix run .# -- --version` to prove the wrapper is self-contained. Verified locally: both steps pass, actionlint and yamllint are clean.

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1238](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1059, opened 2026-07-15
feat(#1016): add CLI option to override lints version via --lints

### LINKED ISSUE #1016: `eoc` CLI missing option to specify `lints` version like it allows for EO parser
Currently, `eoc` allows pinning the EO parser version via `--parser <version>`, which translates to `-Deo.version=...` passed to Maven (`src/mvnw.js:54`). However, there is no way to control which version of [`lints`](https://github.com/objectionary/lints) the `eo-maven-plugin` uses during the `eo:lint` step.

This becomes a problem when testing a fix or a new lint rule from a locally-built or pre-release version of `lints`. The `eo-maven-plugin` picks the `lints` version it was compiled with, and there is currently no way to override it through `eoc`.

**Proposed Solution**

Add a `--lints <version>` option to `eoc`, similar to `--parser`, that passes the corresponding Maven property (e.g. `-Deo.lintsVersion=<version>`) to `mvnw`. The relevant place to thread this through is `src/mvnw.js` (the `flags()` function) and `src/eoc.js` (the option declaration).

**Use Case**

```
eoc --lints 0.0.42 lint
```

This would allow pinning `lints` to a specific version — useful when working across both `eoc` and `lints` repos simultaneously.

Ar2rito:
Closes #1016

## Description

Adds a new `--lints` CLI option to `eoc`, allowing users to specify which version of EO lints to use.

### Problem
Currently, `eoc` allows pinning the parser version via `--parser` but has no equivalent for lints. This makes it impossible to test new or pre-release lint rules.

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -118,6 +118,7 @@ program
   .option('--blind', 'Disable linting')
   .option('--home-tag <version>', 'Git tag in objectionary/home to compile against', tag)
   .option('--parser <version>', 'Set the version of EO parser to use', parser)
+  .option('--lints <version>', 'Set the version of EO lints to use')
   .option('--latest', 'Use the latest parser version from Maven Central')
   .option('--alone', 'Just run a single command without dependencies')
   .option('-l, --language <name>', 'Language of target execution platform (Java or JavaScript, case-insensitive)', language.java)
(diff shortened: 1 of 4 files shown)


Ar2rito:
@yegor256 could you please review this PR?

yegor256:
@Ar2rito how about a test? how do you know this new feature works?

Ar2rito:
@yegor256 check #1098

PR #1068, opened 2026-07-21
#1065: validate --language via commander, catch async command failures

### LINKED ISSUE #1065: `eoc` crashes on invalid `--language` value instead of giving clean error message
## What Happens

An invalid `--language` value crashes `eoc` with a raw, unhandled Node.js stack trace instead of a clean error message.

Reproduced locally:

```
$ node src/eoc.js --language python foreign
Error: Unknown platform python
    at select (src/eoc.js:470:11)
    ... full internal stack trace, including local file paths ...
Node.js v24.15.0
```

The root cause is two separate gaps in `src/eoc.js`:

1. The `select()` function does a manual `throw new Error(\`Unknown platform ${lang}\`)` instead of using commander's built-in `.choices([...])` validation for the `--language` option.
2. `program.parseAsync(process.argv)` (near the bottom of `src/eoc.js`, guarded by `if (require.main === module)`) has no `.catch()` attached. Since command handlers are `async`, any rejected promise anywhere in any command (`parse`, `assemble`, `lint`, `dataize`, `test`, etc.) will crash the process the same way, with a raw stack trace instead of a clean CLI error.

The existing test for this only asserts `status !== 0` on the invalid-language case, so it does not catch that the printed message is an internal stack trace rather than a clean error.

## What Should Happen

Invalid CLI option values should produce a clean, one-line error message and a non-zero exit code, not an internal stack trace. `--language` should validate against the known platform list via commander's `.choices([...])`, and `program.parseAsync(...)` should have a top-level `.catch()` that prints `err.message` and calls `process.exit(1)`, so any future async command failure degrades gracefully instead of leaking internals.

morphqdd:
Closes #1065.

An invalid `--language` value crashed `eoc` with a raw, unhandled Node.js stack trace instead of a clean error message, because `select()` threw a plain `Error` deep inside the async command pipeline, and `program.parseAsync(process.argv)` had no `.catch()` to turn a rejected promise into a clean exit.

Fixed both gaps:

- Added `canonicalLanguage()`, a commander argument processor for the `-l, --language` option. It validates against the same `platforms` map `select()` already used (so `js`, `JavaScript`, `javascript`, etc. all still work, case-insensitively) and throws commander's own `InvalidArgumentError` on anything else. Commander catches that itself and prints a clean one-line message with a non-zero exit, before the value ever reaches the command pipeline.
- Wrapped the top-level `program.parseAsync(process.argv)` call in an async IIFE with try/catch, so any other async command failure prints `err.message` and exits with code 1 instead of an unhandled rejection with a full stack trace.

`select()` itself is unchanged: it is still reachable from other registries in principle, so its own "Unknown platform" check stays as a defensive fallback, even though the `--language` option can no longer feed it an invalid value.

Verified: `npx eslint .` clean. `npx mocha test/test_eoc.js` 15/15 green, including the existing "rejects an unknown --language value" test and the case-insensitive/alias tests. `npx mocha test/commands/test_transpile.js`: the "attempts to transpile a simple .EO program to wrong platform" test passes (checks the same `Unknown platform Eiffel` substring, still present in commander's message); the unrelated "transpiles to Java" test fails the same way on unmodified master too, because this sandbox has no `JAVA_HOME` configured.

Manually confirmed:
```
$ node src/eoc.js --language python foreign
error: option '-l, --language <name>' argument 'python' is invalid. Unknown platform python
$ echo $?
1
```

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -5,7 +5,7 @@
  */
 
 const tinted = require('./tinted-console');
-const {program} = require('commander');
+const {program, InvalidArgumentError} = require('commander');
 
 /**
  * Target language option.
@@ -79,6 +79,22 @@ for (const [alias, canonical] of Object.entries(language)) {
   platforms[canonical.toLowerCase()] = canonical;
 }
 
+/**
+ * Validate and canonicalize the value of the --language option, so an
+ * unknown platform is rejected right away with a clean commander error
+ * instead of surfacing later as an unhandled exception.
+ * @param {String} value - Raw value from the command line
+ * @return {String} The canonical platform name
+ * @throws {InvalidArgumentError} If the language does not resolve to a known platform
+ */
+function canonicalLanguage(value) {
+  const canonical = platforms[String(value).toLowerCase()];
+  if (canonical === undefined) {
+    throw new InvalidArgumentError(`Unknown platform ${value}`);
+  }
+  return canonical;
+}
+
 if (process.argv.includes('--verbose')) {
   tinted.enable('debug');
   console.debug('Debug output is turned ON');
@@ -120,7 +136,7 @@ program
   .option('--parser <version>', 'Set the version of EO parser to use', parser)
   .option('--latest', 'Use the latest parser version from Maven Central')
   .option('--alone', 'Just run a single command without dependencies')
-  .option('-l, --language <name>', 'Language of target execution platform (Java or JavaScript, case-insensitive)', language.java)
+  .option('-l, --language <name>', 'Language of target execution platform (Java or JavaScript, case-insensitive)', canonicalLanguage, language.java)
   .option('-b, --batch', 'Run in batch mode, suppress interactive messages')
   .option('--no-color', 'Disable colorization of console messages')
   .option('--track-transformation-steps', 'Save intermediate XMIR files')
@@ -408,7 +424,14 @@ program.command('fmt')
   });
 
 if (require.main === module) {
-  program.parseAsync(process.argv);
+  (async () => {
+    try {
+      await program.parseAsync(process.argv);
+    } catch (err) {
+      console.error(err.message);
+      process.exit(1);
+    }
+  })();
 }
 
 module.exports.commandsDescription = function commandsDescription() {
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -426,9 +426,20 @@ program.command('fmt')
 if (require.main === module) {
   (async () => {
     try {
+      // @todo #1065:30min Add a test that exercises this catch block through
+      //  a real async command action, not just through commander's own
+      //  InvalidArgumentError handling (the --language=Eiffel case never
+      //  reaches this catch, since commander intercepts it before the
+      //  action runs). Every action that awaits a command properly (parse,
+      //  assemble, transpile, etc.) needs java/mvn/phino to actually fail,
+      //  and the one action that fails without those (generate_comments)
+      //  does not await its own call, so its rejection never reaches here
+      //  either, that is a separate bug worth its own ticket. Fixing that
+      //  bug first (making the action await/return the call) would also
+      //  give this catch block a lightweight, dependency-free test case.
       await program.parseAsync(process.argv);
     } catch (err) {
-      console.error(err.message);
+      console.error(err && err.message ? err.message : err);
       process.exit(1);
     }
   })();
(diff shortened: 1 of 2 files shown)


morphqdd:
@yegor256, please review

Thayorns (changes requested):
Thanks for this — the fix itself is correct and idiomatic. I verified the behavior locally:

```
$ node src/eoc.js --language python foreign
error: option '-l, --language <name>' argument 'python' is invalid. Unknown platform python
$ echo $?
1
```

Clean one-line message, non-zero exit, no stack trace. 👍

My concern is that **this PR ships no tests**, and here that's not a formality — the test gap is part of the bug described in #1065:

> "The existing test for this only asserts `status !== 0` … so it does not catch that the printed message is an internal stack trace rather than a clean error."

The relevant existing test, `test/test_eoc.js:60`, still checks only the exit code:

```js
it('rejects an unknown --language value', (done) => {
  assert.notStrictEqual(eoc('--language=Eiffel', 'clean').status, 0);
```

And `test/commands/test_transpile.js:65` asserts `stderr.includes('Unknown platform Eiffel')` — which a **raw stack trace would also satisfy**. So a regression back to an unhandled stack trace would keep both tests green. The exact behavior this PR fixes is currently unguarded.

Could you please add tests that lock it in:

1. Strengthen "rejects an unknown --language value" to assert stderr **is** the clean commander form (`option '-l, --language <name>' argument … is invalid`) and **does not** contain a stack trace (e.g. no `at select`, no `src/eoc.js:`).
2. Assert `canonicalLanguage` canonicalizes aliases/case (`js` → `JavaScript`, `JAVA` → `Java`).
3. If practical, cover the new `parseAsync` catch path — an async command failure should exit 1 with a clean message rather than an unhandled rejection.

Two smaller notes:

- Validation logic is now duplicated between `canonicalLanguage()` and `select()` (`src/eoc.js:467`). Keeping `select()` as a defensive fallback is reasonable, but consider leaving a `@todo` puzzle rather than a silent copy.
- `console.error(err.message)` in the new catch assumes an `Error`; a non-`Error` rejection would print `undefined`. `err.message ?? err` would be a touch safer.

Also worth flagging: the PR notes the `transpile to Java` test fails locally due to a missing `JAVA_HOME`, so a fully green suite wasn't observed locally — worth confirming CI is green before merge.

As a side benefit, adding these tests also lifts the PR over the 40-hits-of-code floor.

Thayorns (approved):
Re-reviewed after commit `637abcb`. All the concerns from my previous review are addressed, and I ran the updated test suite locally — 19/19 pass, including the 4 language tests.

1. **Strengthened rejection test** — `test/test_eoc.js` now asserts the clean commander message format and the absence of a stack trace (`!stderr.includes('at ')`, `!stderr.includes('eoc.js:')`), so a regression back to a raw stack trace would now fail the test. This closes the exact gap I raised. ✔ (verified passing)
2. **Canonicalization unit tests** — the new `canonicalLanguage` describe block covers `js → JavaScript`, case-insensitive `JavaScript`, `JAVA → Java`, and the throw path. ✔ (verified passing)
3. **parseAsync catch path** — deferred with a well-reasoned `@todo #1065:30min` puzzle explaining why it is hard to test dependency-free (commander intercepts `InvalidArgumentError` before the action; the real async actions need java/mvn/phino to fail; `generate_comments` does not await its own call — a separate latent bug worth its own ticket). Reasonable Puzzle-Driven deferral, and the catch was also hardened to `err && err.message ? err.message : err`. ✔
4. **Duplicated validation in select()** — documented with an `@todo #1065:30min` puzzle and a concrete dedupe plan, rather than a silent copy. ✔

Size is now ~85 hits-of-code, within guidelines.

Note: I could not re-run `npx eslint` in my local environment (missing devDependency locally, unrelated to this PR) — please confirm lint is green in CI, but behavior and tests are verified. LGTM.

morphqdd:
@yegor256, review pls

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+229](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1361](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1069, opened 2026-07-21
#1064: refuse to delete cwd, its ancestors, or the home directory

### LINKED ISSUE #1064: `eoc clean` command risks deleting crucial directories without prompt or confirmation
## What Happens

`src/commands/clean.js` deletes the directory given via `--target` with `fs.rmSync(home, {recursive: true, force: true})`, where `home = path.resolve(opts.target)`. There is no check that the resolved path is not the current working directory, an ancestor of it, or the user's home directory.

Reproduced locally:

```
$ mkdir eoc-clean-test && cd eoc-clean-test && touch important-file.txt myproject.eo
$ node path/to/eoc.js --target . clean
The directory . was deleted
$ ls -la
# the directory and everything in it is gone; pwd fails afterward
```

The same happens with `--target ..`: it deletes the parent directory, destroying its siblings and the directory the command was run from.

A typo in `--target` (or a wrapper script that computes `--target`/`--dir` incorrectly) silently wipes a project, or the parent of a project, with no prompt and no dry-run option.

## What Should Happen

`eoc clean` should refuse to delete when the resolved target path equals `process.cwd()`, is an ancestor of `process.cwd()`, or equals the user's home directory (`os.homedir()`). At minimum it should require an explicit confirmation or a `--force` flag before deleting anything that is not clearly a project-local temp directory (the documented default is `.eoc`).

morphqdd:
Closes #1064.

`eoc clean --target <path>` deleted the resolved target with `fs.rmSync(home, {recursive: true, force: true})` with no check that the path was not the current working directory, an ancestor of it, or the user's home directory. A typo in `--target` (for example `.` instead of the documented default `.eoc`) silently wiped a project or its parent with no prompt.

Added `isSafeToDelete()` in `src/commands/clean.js`: it refuses (throws) when the resolved target equals `process.cwd()`, is an ancestor of it (checked via `path.relative`, so it works for any depth), or equals `os.homedir()`. This matches the existing codebase convention of command modules throwing a plain `Error` for validation failures (see `normalize.js`, `lint.js`, `docs.js`).

Three new tests in `test/commands/test_clean.js` cover the three refusal cases; the two existing tests (delete with/without `--global`) still pass unchanged, confirming the normal `.eoc`-subdirectory case is unaffected.

Verified: `npx eslint .` clean. `npx mocha test/commands/test_clean.js` 5/5 green. `npx mocha test/test_eoc.js` 15/15 green (no regressions).

Manually confirmed:
```
$ mkdir eoc-clean-test && cd eoc-clean-test && touch important-file.txt
$ node path/to/eoc.js --target . clean
Error: Refusing to delete .: it is the current directory, an ancestor of it, or the home directory
$ ls
important-file.txt
```

### DIFF
--- a/src/commands/clean.js
+++ b/src/commands/clean.js
@@ -9,19 +9,19 @@ const path = require('path');
 const os = require('os');
 
 /**
- * Checks that deleting the given directory would not destroy anything
- * outside the project's own temporary files: the current working
- * directory itself, an ancestor of it, or the user's home directory.
+ * Refuses, by throwing, to delete a directory that is the current
+ * working directory, an ancestor of it, or the user's home directory.
  * @param {String} target - Resolved absolute path of the directory to delete
- * @return {Boolean} True when the target is not the current directory,
- *  an ancestor of it, or the home directory
  */
-function isSafeToDelete(target) {
+function guard(target) {
   const cwd = process.cwd();
-  const fromTargetToCwd = path.relative(target, cwd);
-  const targetIsCwdOrAncestor = fromTargetToCwd === '' ||
-    (!fromTargetToCwd.startsWith('..') && !path.isAbsolute(fromTargetToCwd));
-  return !targetIsCwdOrAncestor && target !== os.homedir();
+  const route = path.relative(target, cwd);
+  const encloses = route === '' || (!route.startsWith('..') && !path.isAbsolute(route));
+  if (encloses || target === os.homedir()) {
+    throw new Error(
+      `Refusing to delete ${rel(target)}: it is the current directory, an ancestor of it, or the home directory`
+    );
+  }
 }
 
 /**
@@ -30,11 +30,7 @@ function isSafeToDelete(target) {
  */
 module.exports = function(opts) {
   const home = path.resolve(opts.target);
-  if (!isSafeToDelete(home)) {
-    throw new Error(
-      `Refusing to delete ${rel(home)}: it is the current directory, an ancestor of it, or the home directory`
-    );
-  }
+  guard(home);
   if (fs.existsSync(home)) {
     fs.rmSync(home, {recursive: true, force: true});
     console.info('The directory %s was deleted', rel(home));
--- a/src/commands/clean.js
+++ b/src/commands/clean.js
@@ -10,10 +10,13 @@ const os = require('os');
 
 /**
  * Refuses, by throwing, to delete a directory that is the current
- * working directory, an ancestor of it, or the user's home directory.
+ * working directory, an ancestor of it, or the user's home directory;
+ * otherwise returns it unchanged, so the guard cannot be skipped or
+ * reordered away from the value it protects.
  * @param {String} target - Resolved absolute path of the directory to delete
+ * @return {String} The same target, once confirmed safe to delete
  */
-function guard(target) {
+function guarded(target) {
   const cwd = process.cwd();
   const route = path.relative(target, cwd);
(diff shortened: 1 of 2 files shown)


morphqdd:
@yegor256, please review

Thayorns (changes requested):
Thanks — the behavior is correct. I verified the guard across the cwd / ancestor / normal `.eoc` subdir / unrelated sibling / homedir cases and it refuses exactly the dangerous ones per #1064, the three new tests cover the refusal paths, and it stays within the size guidelines. My only issue is naming.

CLAUDE.md is explicit: *"Variables: single nouns" / "Methods: single verbs" / "Avoid compound variable names, maximum 2 words per variable."* Three of the new names go well past that:

| Name | Words |
|---|---|
| `isSafeToDelete` | 4 |
| `fromTargetToCwd` | 4 |
| `targetIsCwdOrAncestor` | 4–5 |

The cleaner modules in this repo set the bar the other way — `lint.js` names helpers `goals()` / `extras()`, and `normalize.js` uses single-noun locals (`sources`, `target`, `parsed`, `normed`).

Beyond the composite names, the boolean helper also forces a double negative at the call site (`if (!isSafeToDelete(home)) throw`). Collapsing it into a throwing **guard** (single verb, fail-fast — which the project favors) removes both the composite name and the negation, and keeps the message out of the caller:

```js
/**
 * Refuses, by throwing, to delete a directory that is the current
 * working directory, an ancestor of it, or the user's home directory.
 * @param {String} target - Resolved absolute path of the directory to delete
 */
function guard(target) {
  const cwd = process.cwd();
  const route = path.relative(target, cwd);
  const encloses = route === '' || (!route.startsWith('..') && !path.isAbsolute(route));
  if (encloses || target === os.homedir()) {
    throw new Error(
      `Refusing to delete ${rel(target)}: it is the current directory, an ancestor of it, or the home directory`
    );
  }
}
```

Call site becomes just `guard(home);`. Renames: `isSafeToDelete → guard`, `fromTargetToCwd → route`, `targetIsCwdOrAncestor → encloses` — all ≤2 words, no double negative.

If you would rather keep the minimal change, at least rename the three composites; but the guard form is cleaner and matches the "fail fast" convention.

Thayorns (approved):
Looks good!

src/commands/clean.js
```diff
@@ -8,12 +8,29 @@ const fs = require('fs');
...
+  }
+}
+
 /**
  * Deletes all temporary files.
  * @param {Hash} opts - All options
  */
 module.exports = function(opts) {
   const home = path.resolve(opts.target);
+  guard(home);
```
yegor256: @morphqdd read this: https://www.yegor256.com/2015/12/08/temporal-coupling-between-method-calls

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+225](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1509](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1075, opened 2026-07-22
fix(#1071): await generate_comments action so failures reach the top-level catch

### LINKED ISSUE #1071: Uncovered catch block in eoc.js requires test coverage
The puzzle `1065-f9257add` from #1065 has to be resolved: 

https://github.com/objectionary/eoc/blob/4998dee3e1bceae51476524fd7eebdb67377ac22/src/eoc.js#L429-L439

The puzzle was created by @insomnia on 21-Jul-26. 

Thayorns:
The `generate_comments` command action called its async command without awaiting it, so any failure escaped as an unhandled promise rejection with a full stack trace instead of the clean, one-line message the top-level `catch` in `eoc.js` is meant to print. That catch block therefore had no path that reached it from a real command, which is what the #1065 puzzle flagged.

Awaiting the call inside the action lets the rejection propagate through `parseAsync` into the catch, so a failing command now exits with code 1 and a single readable line.

Two tests cover the catch through distinct error types — a custom `Error` from an unknown `--provider`, and a Node `ENOENT` from a missing template file — each asserting a non-zero exit and the absence of any `.js:line` stack frames. Both need no Java, Maven, or network, so they stay fast and hermetic. The resolved puzzle text is removed.

Closes #1071.

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -352,8 +352,8 @@ program.command('generate_comments')
   .requiredOption('--prompt_template <path>',
     'Path to prompt template file, ' +
     'where `{code}` placeholder will be replaced with the code given by the user')
-  .action((str, opts) => {
-    coms().generate_comments({...program.opts(), ...str});
+  .action(async (str, opts) => {
+    await coms().generate_comments({...program.opts(), ...str});
   });
 
 program.command('jeo:disassemble')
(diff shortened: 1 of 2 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+185](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1077, opened 2026-07-22
#1073: await command action handlers in eoc.js so failures reach the top-level catch

### LINKED ISSUE #1073: Unawaited Promises in Several Command Action Handlers in `eoc.js` Cause Crash
## What Happens

Several command action handlers in `src/eoc.js` call their command function without `await` or `return`, even though every one of those command modules returns a Promise. The affected handlers are:

- `audit` (calls `src/commands/audit.js`)
- `register` (calls `src/commands/register.js`)
- `print` (calls `src/commands/print.js`)
- `docs` (calls `src/commands/docs.js`)
- `generate_comments` (calls `src/commands/generate_comments.js`)
- `jeo:disassemble` (calls `src/commands/jeo/disassemble.js`)
- `jeo:assemble` (calls `src/commands/jeo/assemble.js`)

All other pipeline commands (`parse`, `assemble`, `lint`, `resolve`, `transpile`, `compile`, `link`, `dataize`, `test`, `latex`, `normalize`, `fmt`) correctly declare their action as `async` and `await` the work.

Because these 7 handlers do not await the promise, a real failure inside the command never reaches the top-level `try { await program.parseAsync(...) } catch (err) { console.error(err.message); process.exit(1); }` block in `eoc.js`. Instead it surfaces as a raw, unhandled Node.js stack trace.

Reproduction 1, `generate_comments` with an unsupported provider:

```
node src/eoc.js generate_comments --provider bogus --source src/eoc.js --prompt_template src/eoc.js
```

Prints a full Node stack trace ending in `Node.js v24.15.0` and exits 1, instead of a clean one-line error.

Reproduction 2, `docs` with a `--target` that points at a file instead of a directory:

```
touch notadir
node src/eoc.js --target notadir docs
```

Prints an internal `console.error('Error generating documentation:', ...)` log, then crashes a second time with a separate raw `ENOTDIR` stack trace from `node:internal/process/promises`.

For comparison, an option-validation error already produces the clean behavior this should have:

```
node src/eoc.js --language Eiffel parse
```
prints only: `error: option '-l, --language <name>' argument 'Eiffel' is invalid. Unknown platform Eiffel`

## What Should Happen

Each of the 7 action handlers listed above should be declared `async` and should `await`/`return` the promise from its command function, the same way every other pipeline command already does. That way a real failure in any of them produces the same clean one-line error and exit code 1, instead of a raw stack trace.

morphqdd:
Closes #1073.

Seven command action handlers in `src/eoc.js` (`audit`, `register`, `print`, `docs`, `generate_comments`, `jeo:disassemble`, `jeo:assemble`) called their command function without `await` or `return`, even though every one of those command modules returns a Promise. A real failure inside any of them never reached the top-level `try { await program.parseAsync(...) } catch (err) { ... } ` block, so it surfaced as a raw, unhandled Node.js stack trace instead of the clean one-line error every other pipeline command already produces.

Fix: each of the 7 handlers is now declared `async` and awaits its command's promise, matching the idiom already used by `parse`, `assemble`, `transpile`, and the rest.

Also removed the `@todo #1065:30min ...` puzzle comment above the top-level catch block. That puzzle asked for a test that exercises the catch block through a real async action failure (not just commander's own option-validation path), which this fix now makes possible and which the two new tests below provide.

Added two tests to the existing `describe('eoc', ...)` block in `test/test_eoc.js` (reusing its existing `eoc(...args)` CLI-spawning helper rather than adding a new block):
1. `generate_comments --provider=bogus` now exits non-zero with a single clean line, no stack trace.
2. `docs` with a `--target` pointing at a file (not a directory) now exits non-zero with a clean `ENOTDIR: not a directory, mkdir ...` line as the last line of stderr, instead of crashing with an unhandled rejection.

Note on the `docs` test: `docs.js` has its own separate, pre-existing wart where it logs an internal stack dump before rethrowing. That is a distinct bug, out of scope here, so the test checks that the clean message is the *last* line reaching the top-level catch rather than asserting the whole of stderr is stack-trace-free.

Note for reviewers: another open PR (#1075, for issue #1071) also touches the `generate_comments` handler and removes the same `@todo #1065` puzzle block, since fixing this bug is what its puzzle was actually asking for. Whichever of the two merges first, the other will need a small rebase.

Verified:
- `npx mocha test/test_eoc.js`: 21/21 passing.
- `npx eslint .`: clean.
- Manually reproduced both cases from the issue: `generate_comments --provider bogus` and `docs` with a file `--target`, both now exit 1 with a clean message instead of a raw stack trace.

### DIFF
--- a/test/test_eoc.js
+++ b/test/test_eoc.js
@@ -69,7 +69,53 @@ describe('eoc', () => {
     assert(!stderr.includes('eoc.js:'), stderr);
     done();
   });
+  it('reports a clean error when generate_comments gets an unsupported provider, instead of a raw stack trace', (done) => {
+    const result = eoc(
+      'generate_comments',
+      '--provider=bogus',
+      '--source', path.resolve('./src/eoc.js'),
+      '--prompt_template', path.resolve('./src/eoc.js')
+    );
+    const stderr = result.stderr.toString();
+    assert.notStrictEqual(result.status, 0);
+    assert(!stderr.includes('Node.js v'), stderr);
+    assert(!stderr.includes('node:internal/process/promises'), stderr);
+    assert(!stderr.includes('at '), stderr);
+    assert.strictEqual(
+      lastLine(stderr),
+      '`bogus` provider is not supported. Currently supported providers are: `openai`, `placeholder`'
+    );
+    done();
+  });
+  it('reports a clean error when docs gets a --target that is a file, instead of an unhandled-rejection crash', (done) => {
+    const fs = require('fs'),
+      os = require('os'),
+      home = fs.mkdtempSync(path.join(os.tmpdir(), 'eoc-docs-target-')),
+      target = path.join(home, 'notadir');
+    fs.writeFileSync(target, '');
+    const result = eoc('--target', target, 'docs');
+    const stderr = result.stderr.toString();
+    assert.notStrictEqual(result.status, 0);
+    assert(!stderr.includes('Node.js v'), stderr);
+    assert(!stderr.includes('node:internal/process/promises'), stderr);
+    assert(/^ENOTDIR: not a directory, mkdir/.test(lastLine(stderr)), stderr);
+    done();
+  });
 });
+
+/**
+ * Extract the last non-empty line of a stderr string, stripped of ANSI
+ * color codes. This is where the top-level
+ * `program.parseAsync(...).catch(...)` block in eoc.js prints a rejected
+ * command's clean error message.
+ * @param {String} stderr - Raw stderr text
+ * @return {String} The last non-empty line
+ */
+function lastLine(stderr) {
+  const lines = stderr.split('\n').map((l) => l.trim()).filter((l) => l.length > 0);
+  // eslint-disable-next-line no-control-regex
+  return lines[lines.length - 1].replace(/\x1b\[[0-9;]*m/g, '');
+}
 
 describe('canonicalLanguage', () => {
   const {canonicalLanguage} = require('../src/eoc');
--- a/test/test_eoc.js
+++ b/test/test_eoc.js
@@ -112,9 +112,10 @@ describe('eoc', () => {
  * @return {String} The last non-empty line
  */
 function lastLine(stderr) {
-  const lines = stderr.split('\n').map((l) => l.trim()).filter((l) => l.length > 0);
-  // eslint-disable-next-line no-control-regex
-  return lines[lines.length - 1].replace(/\x1b\[[0-9;]*m/g, '');
+  const esc = String.fromCharCode(27),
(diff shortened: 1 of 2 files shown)


test/test_eoc.js
```diff
@@ -69,8 +69,54 @@ describe('eoc', () => {
...
+ * Extract the last non-empty line of a stderr string, stripped of ANSI
+ * color codes. This is where the top-level
+ * `program.parseAsync(...).catch(...)` block in eoc.js prints a rejected
+ * command's clean error message.
+ * @param {String} stderr - Raw stderr text
+ * @return {String} The last non-empty line
+ */
+function lastLine(stderr) {
+  const lines = stderr.split('\n').map((l) => l.trim()).filter((l) => l.length > 0);
+  // eslint-disable-next-line no-control-regex
```
Thayorns: Please avoid disabling eslint rules unless it is genuinely unavoidable. This `no-control-regex` disable is not necessary here: the ANSI stripping can be expressed so the escape character is not a control-character literal inside a regex, and the rule stays satisfied with no disable. Could you refactor the helper so this line can go away?

Thayorns (changes requested):
The fix itself looks correct and complete: the seven handlers that return promises are now awaited, and `foreign`/`clean` are synchronous, so leaving those un-awaited is right. The catch-block tests are a good, dependency-free addition.

One request: please avoid disabling eslint rules where it is not necessary. The `no-control-regex` disable in the new `lastLine` helper is avoidable while keeping the exact same behaviour. See the inline comment.

morphqdd:
@Thayorns thanks for the review. Fixed: `lastLine` now builds the ANSI-strip pattern at runtime via `String.fromCharCode(27)` and `new RegExp(...)`, so there is no literal control character in the regex source and the `no-control-regex` disable is gone. Same behavior, verified with `npx eslint` (clean) and `npx mocha test/test_eoc.js` (21/21 passing).

Thayorns (approved):
Looks solid

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+224](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1728](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1078, opened 2026-07-22
fix(#1072): remove duplicated unknown-platform check from select()

### LINKED ISSUE #1072: Duplicated 'unknown platform' in eoc.js needs removal
The puzzle `1065-e9c76f78` from #1065 has to be resolved: 

https://github.com/objectionary/eoc/blob/4998dee3e1bceae51476524fd7eebdb67377ac22/src/eoc.js#L498-L507

The puzzle was created by @insomnia on 21-Jul-26. 

Thayorns:
`select()` was repeating the exact platform lookup and `Unknown platform` throw that `canonicalLanguage()` already does. Since every language reaching `select()` has been through `canonicalLanguage()` (the commander coercion for `--language`), the second copy was unreachable from the CLI and only risked drifting from the first.

`select()` now resolves the registry with `registry[canonicalLanguage(lang)]`, so the validation and the `Unknown platform` message live in a single place. Behaviour is unchanged: aliases and mixed case still resolve, and an unknown platform still fails fast, now through `canonicalLanguage()`.

`select()` is exported and covered by two tests: one checks that an alias resolves to the canonical registry entry, the other that an unregistered platform throws. The existing `--language` and wrong-platform tests still pass.

Closes #1072.

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -70,8 +70,8 @@ const pipelines = {
  * canonical platform key. Derived from `language` so the accepted names
  * and the platforms can never drift apart. For each `[alias, canonical]`
  * pair we accept the short alias (e.g. `js`) and the canonical name
- * lower-cased (e.g. `javascript`); `select()` lower-cases the input, so
- * mixed-case spellings such as `JavaScript` are matched too.
+ * lower-cased (e.g. `javascript`); `canonicalLanguage()` lower-cases the
+ * input, so mixed-case spellings such as `JavaScript` are matched too.
  */
 const platforms = {};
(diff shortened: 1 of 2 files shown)


0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+201](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1081, opened 2026-07-22
fix(#1074): align Java test command execution

### LINKED ISSUE #1074: `test.js` inconsistently passes args to `mvnw` and doesn't use `elapsed()` wrapper like other commands
## What Happens

`src/commands/java/test.js` calls the Maven wrapper differently from every one of its sibling pipeline commands.

Line 30:
```js
return maven(args.concat(flags(opts)));
```

Compare with `resolve.js`, `transpile.js`, `compile.js`, `parse.js`, `assemble.js`, `lint.js`, `normalize.js`, `latex.js`, all of which call:
```js
return elapsed(async (tracked) => {
  const r = await mvnw(goals(...).concat(flags(opts)), opts.target, opts.batch);
  tracked.print(...);
  return r;
});
```

`test.js` is the only pipeline command that:
1. Does not pass `opts.target` and `opts.batch` to the Maven call.
2. Is not wrapped in `elapsed(...)`.

Because `mvnw(args, tgt, batch)` gates its interactive progress spinner on `tgt !== undefined`, `eoc test` never shows the progress indicator that `resolve`/`transpile`/`compile` show in interactive (non `--batch`) mode. Because there is no `elapsed()` wrapper, it also never prints the "done in Xs" summary line the other commands print.

The existing test for this command, `test/commands/java/test_test.js`, only checks a single-argument callback (`(args) => { captured = args; }`), so it cannot catch that `target` and `batch` are missing from the call — the test coverage gap matches the bug exactly.

This is not a new pattern to this repo: issues #434/#435 already established that pipeline commands should be wrapped in `elapsed()` with a print summary, fixing `compile.js` for it in 2024. `test.js` was simply never brought in line with that pattern.

## What Should Happen

`test.js` should match its siblings:
```js
return elapsed(async (tracked) => {
  const r = await maven(args.concat(flags(opts)), opts.target, opts.batch);
  tracked.print(...);
  return r;
});
```

Mahan1341:
Closes #1074.

The Java `test` command previously called the Maven runner without passing
`opts.target` and `opts.batch`. It was also the only Java pipeline command
not wrapped with `elapsed(...)`.

This change:

- passes the target directory and batch mode to the Maven runner;
- wraps test execution with `elapsed(...)`;
- prints the execution duration after the tests finish;
- adds a regression test for the execution options and timing output.

Tests:

- `npx mocha test/commands/java/test_test.js` — 4 passing
- `npx grunt` — 135 passing, 10 pending

### DIFF
--- a/test/commands/java/test_test.js
+++ b/test/commands/java/test_test.js
@@ -39,4 +39,21 @@ describe('java/test', () => {
       `expected no -Dtest arg, got: ${captured}`
     );
   });
+  it('passes execution options to Maven', async () => {
+    let captured;
+    await test(
+      {
+        stack: '64M',
+        heap: '256M',
+        sources: 'src',
+        target: 'target',
+        batch: true,
+      },
+      (args, target, batch) => {
+        captured = {args, target, batch};
+      }
+    );
+    assert.strictEqual(captured.target, 'target');
+    assert.strictEqual(captured.batch, true);
+  });
 });
--- a/src/commands/java/test.js
+++ b/src/commands/java/test.js
@@ -4,6 +4,7 @@
  */
 
 const {mvnw, flags} = require('../../mvnw');
+const {elapsed} = require('../../elapsed');
 const {verifyJavac} = require('../../jdk');
 
 /**
@@ -27,5 +28,13 @@ module.exports = function(opts, maven = mvnw) {
     const cls = `EO${obj}*Test`;
     args.push(`-Dtest=${pkg ? `org.eolang.${pkg}.${cls}` : `org.eolang.${cls}`}#${method}`);
   }
-  return maven(args.concat(flags(opts)));
+  return elapsed(async (tracked) => {
+    const result = await maven(
+      args.concat(flags(opts)),
+      opts.target,
+      opts.batch
+    );
+    tracked.print('Java tests completed');
+    return result;
+  });
 };


test/commands/java/test_test.js
```diff
@@ -39,4 +39,35 @@ describe('java/test', () => {
...
+        }
+      );
+    } finally {
+      console.info = original;
+    }
+    assert.strictEqual(captured.target, 'target');
+    assert.strictEqual(captured.batch, true);
+    assert.match(
+      logged,
+      /^Java tests completed in \d+(ms|s|min)$/
```
Thayorns: This overrides the global `console.info` and asserts on the captured line — i.e. it tests a log, which the project bylaws rule out.

Thayorns (changes requested):
The source change is right: passing `target`/`batch` to the Maven runner and wrapping in `elapsed(...)` matches the other Java commands, so it correctly resolves #1074. (The `console.info` there comes from `elapsed` — standard pattern, fine to keep.)

The problem is the new test — see the inline note on the assertion. Separately, that timing check is redundant: the `in <duration>` formatting belongs to `elapsed` and is already covered by `test/test_elapsed.js`, so only the `target`/`batch` assertions here earn their keep. There's also a stray blank line inside the `describe` block.

test/commands/java/test_test.js
Mahan1341: Fixed, thank you. I removed the global `console.info` override and the redundant elapsed-time assertion. The regression test now checks only that `target` and `batch` are passed to the Maven runner.

Mahan1341:
@Thayorns, thank you. I pushed the requested test cleanup and resolved the conversation. Could you please re-review the changes?

test/commands/java/test_test.js
```diff
@@ -39,4 +39,21 @@ describe('java/test', () => {
...
+        sources: 'src',
+        target: 'target',
+        batch: true,
+      },
+      (args, target, batch) => {
+        captured = {args, target, batch};
+      }
+    );
+    assert.strictEqual(captured.target, 'target');
+    assert.strictEqual(captured.batch, true);
```
Thayorns, line 57: LGTM

Thayorns (approved):
Looks solid

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (3) comments; -4 for too few (28) hits-of-code; +1 to give you at least something.  Your running score is [+235](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Mahan1341:
@yegor256, please review. The only failing check is `xcop` on `mvnw/pom.xml`, which is not modified in this PR.

0crat:
@mahan1341 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+48](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1090, opened 2026-07-23
fix: print resolved paths in jeo commands (#1088)

### LINKED ISSUE #1088: `jeo:assemble`/`jeo:disassemble` print the wrong (unresolved) path in their success message
## What Happens

`src/commands/jeo/assemble.js` and `src/commands/jeo/disassemble.js` (lines 30-32 in each) resolve `--xmirs`/`--classes` against `--target` for the actual Maven flags:

```js
`-Djeo.assemble.sourcesDir=${path.resolve(opts.target, opts.xmirs)}`,
`-Djeo.assemble.outputDir=${path.resolve(opts.target, opts.classes)}`,
```

but the success message printed afterward uses the raw, unresolved options instead of those same resolved paths:

```js
tracked.print(`EO .xmir files from ${rel(opts.xmirs)} assembled to .class files in ${rel(opts.classes)}`);
```

`--xmirs`/`--classes` are documented as "relative to --target" (`src/eoc.js:364-389`). The actual file operation happens under `--target`, but the printed confirmation shows the raw option value interpreted relative to the current working directory, which is wrong whenever `--target` differs from cwd. Every other command in this repo (`assemble.js`, `parse.js`, `transpile.js`, `compile.js`, etc.) prints a path built from `path.resolve(opts.target, ...)`, so the two `jeo` commands are the odd ones out.

Verified by stubbing `mvnw` and calling `jeo/assemble.js` directly with `target: '/tmp/mytarget', xmirs: 'xmir', classes: 'classes'`:

```
mvnw goals: -Djeo.assemble.sourcesDir=/tmp/mytarget/xmir -Djeo.assemble.outputDir=/tmp/mytarget/classes
PRINTED MESSAGE: EO .xmir files from xmir assembled to .class files in classes
```

The printed message never mentions `/tmp/mytarget` at all. A user running from a different working directory would look in the wrong place for the output. Existing tests (`test/commands/jeo/test_assemble.js`) mask this because they always pass `--xmirs`/`--classes` as absolute paths, so `path.resolve` ignores `--target` and the bug never triggers.

## What Should Happen

The success message in both commands should print `rel(path.resolve(opts.target, opts.xmirs))` and `rel(path.resolve(opts.target, opts.classes))`, matching the paths actually passed to Maven.

Owwleet:
Fixes #1088
Changed the success message in `assemble.js` and `disassemble.js` to use
`path.resolve(opts.target, opts.xmirs)` and `path.resolve(opts.target, opts.classes)`
instead of the raw option values. This ensures the printed paths match
the actual paths passed to Maven.

Before:
```js
tracked.print(`EO .xmir files from ${rel(opts.xmirs)} assembled to .class files in ${rel(opts.classes)}`);

### DIFF
--- a/src/commands/jeo/assemble.js
+++ b/src/commands/jeo/assemble.js
@@ -27,9 +27,7 @@ module.exports = function(opts) {
         ),
       opts.target, opts.batch
     );
-    tracked.print(
-      `EO .xmir files from ${rel(opts.xmirs)} assembled to .class files in ${rel(opts.classes)}`
-    );
+    tracked.print(`EO .xmir files from ${rel(path.resolve(opts.target, opts.xmirs))} assembled to .class files in ${rel(path.resolve(opts.target, opts.classes))}`);
     return r;
   });
 };
--- a/src/commands/jeo/disassemble.js
+++ b/src/commands/jeo/disassemble.js
@@ -27,9 +27,7 @@ module.exports = function(opts) {
         ),
       opts.target, opts.batch
     );
-    tracked.print(
-      `Bytecode .class files from ${rel(opts.classes)} disassembled to .xmir files in ${rel(opts.xmirs)}`
-    );
+    tracked.print(`Bytecode .class files from ${rel(path.resolve(opts.target, opts.classes))} disassembled to .xmir files in ${rel(path.resolve(opts.target, opts.xmirs))}`);
     return r;
   });
 };


0crat:
@owwleet Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1093, opened 2026-07-24
fix(#1083): group docs packages in a Map so prototype-named packages do not crash

### LINKED ISSUE #1083: `docs` command fails when generating documentation for package named `constructor`
# `docs` crashes for the `constructor` package

## What Happens

`eoc docs` crashes when parser-produced XMIR belongs to an EO package named `constructor`.

Minimal EO source at `src/constructor/foo.eo`:

```eo
+package constructor

# Foo.
[] > foo
```

The `parse` command succeeds and creates `target/1-parse/constructor/foo.xmir`. The generated XMIR contains `<tail>constructor</tail>` for the package meta. Running `docs` on the same target exits with code 1:

```text
Error generating documentation: TypeError: Cannot read properties of undefined (reading 'push')
    at src/commands/docs.js:151:44
...
packages_info[packages].xmir_htmls.push(xmir_html);
                             ^

TypeError: Cannot read properties of undefined (reading 'push')
```

## Steps to Reproduce

On commit `ddd1245c6f474e7bee50238d64072a9e8d853bb8`, create the EO source above and run:

```bash
node src/eoc.js --batch \
  --sources /tmp/eoc-docs-constructor/src \
  --target /tmp/eoc-docs-constructor/target \
  parse

node src/eoc.js --batch \
  --sources /tmp/eoc-docs-constructor/src \
  --target /tmp/eoc-docs-constructor/target \
  docs
```

Actual result:

- `parse`: exit code 0; `Pipeline [register → parse] done`
- `docs`: exit code 1 with the stack trace above

Control case: changing only the package name to `ordinary` makes `docs` exit 0 and generate:

```text
docs/ordinary/foo.html
docs/package_ordinary.html
docs/packages.html
docs/styles.css
docs/summary.xml
```

The failure was reproduced both on Node `18.20.8` (the version in the project's `grunt` and integration-test matrices) and Node `26.4.0`. The control case passed on both versions.

## What Should Happen

`docs` should generate the object page, package page, package index, CSS, and summary XML for the parser-accepted `constructor` package, just as it does for `ordinary`, and exit with code 0.

## Environment

- `objectionary/eoc` commit: `ddd1245c6f474e7bee50238d64072a9e8d853bb8` (`master`)
- Node: `18.20.8` and `26.4.0`
- npm: `11.17.0` for the baseline installation
- EO parser: `0.61.3` (`392dae0`)
- OS: macOS `26.5.2`, arm64
- Baseline: `npm test` — 134 passing, 10 pending; `npx eslint .` — exit code 0

## Relevant Code and Confirmed Cause

`src/commands/docs.js:132-151` stores package data in a normal object and checks membership with `in`:

```js
const packages_info = {};
// ...
if (!(packages in packages_info)) {
  packages_info[packages] = {
    xmir_htmls : [],
    names: [],
    path: html_package
  };
}
packages_info[packages].xmir_htmls.push(xmir_html);
```

For `packages === 'constructor'`, `'constructor' in packages_info` is already `true` because the property is inherited from `Object.prototype`. Initialization is skipped; the inherited constructor function has no `xmir_htmls`, so line 151 dereferences `undefined`.

## Duplicate Check

Searched open and closed issues and all PRs for `docs constructor`, `constructor package`, `prototype package docs`, `packages_info`, `reading 'push'`, `docs crash package`, and the exact exception text.

The closest results are not duplicates:

- #467 / PR #951 add `summary.xml` generation; their scope is missing output, not package-name-dependent failure. The same object-membership check predates that PR.
- #1073 covers unawaited command promises and explains why failures may print a raw stack twice; it does not cover or prevent the package-name crash itself.
- #253 covers an invalid parser version and an unrelated Maven error.

No open or closed issue or PR describing a `constructor` package collision in `docs` was found.

## Acceptance Criteria

- The EO example above completes both `parse` and `docs` with exit code 0.
- Documentation artifacts are generated for the `constructor` package, including `constructor/foo.html`, its package page, `packages.html`, and `summary.xml`.
- Existing ordinary package documentation remains unchanged.
- A regression test uses a package key inherited by a normal JavaScript object, at minimum `constructor`.

Thayorns:
The `docs` command grouped packages in a plain object and checked membership with the `in` operator, which walks the prototype chain. A package named after an inherited property — `constructor`, and likewise `toString`, `valueOf`, `hasOwnProperty` — was treated as already present, so its entry was never initialized and pushing to `xmir_htmls` dereferenced `undefined`, crashing the command with `Cannot read properties of undefined (reading 'push')`.

Switching the grouping to a `Map` keys on the string directly, so every package name is handled regardless of what it collides with on `Object.prototype`, including `__proto__` (where a plain-object assignment would otherwise set the prototype instead of a key).

Added a regression test that runs `docs` on a `constructor` package and asserts the package page is generated.

Verified: `npx mocha test/commands/test_docs.js` 10/10 passing, `npx eslint .` clean.

Closes #1083.

### DIFF
--- a/src/commands/docs.js
+++ b/src/commands/docs.js
@@ -129,7 +129,7 @@ module.exports = function(opts) {
       fs.mkdirSync(output, {recursive: true});
       const css = path.join(output, 'styles.css');
       fs.writeFileSync(css, '');
-      const packages_info = {};
+      const packages_info = new Map();
       const all_xmir_htmls = [];
       const xmirs = findFiles(input, '.xmir');
       for (const xmir of xmirs) {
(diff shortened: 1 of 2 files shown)


Thayorns:
@yegor256 take a look please.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+217](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1094, opened 2026-07-24
fix(#1085): only treat a 404 as a confirmed-absent parser version

### LINKED ISSUE #1085: `parser-version.js` incorrectly marks versions as absent for transient Maven Central errors (503/429), not just confirmed-absent 404
## What Happens

Issue #998 ("`eoc` outputs incorrect 'not available in Maven Central' error during network unavailability") was fixed by PR #1020: when `fetch()` throws (DNS failure, connection refused, a timeout that throws), `src/parser-version.js`'s `exists()` function now correctly logs a warning and assumes the version exists, instead of hard-failing.

That fix only covers exceptions. `exists()` (line 51) still does:

```js
result = res.statusCode === 200;
```

When Maven Central responds with a non-200, non-404 status (503 Service Unavailable, 502 Bad Gateway, 429 rate-limited, all common during outages or throttling), `fetch()` does not throw, it resolves, and this line evaluates to `false`, exactly the same result as a confirmed 404.

That `false` propagates to `src/mvnw.js` (lines 56-63), which prints "Parser version X is not available in Maven Central" and calls `process.exit(1)`, the very false-negative issue #998 was filed to eliminate, just reached through the response path instead of the exception path.

Verified directly:

```
$ node -e "... parserVersion.exists('0.28.11', () => ({statusCode: 503})) ..."
statusCode 503 (transient server error) -> exists(): false
statusCode 429 (rate limited)          -> exists(): false
statusCode 404 (confirmed absent)      -> exists(): false
thrown network error (ECONNRESET)      -> exists(): true   <- correctly lenient
```

503/429 (transient) and 404 (confirmed absent) all return `false` identically, while an actual thrown exception is correctly lenient. PR #1020's own added tests only cover status codes `200` and `404`; 5xx/429 were never tested or fixed.

A perfectly valid parser version, even the bundled default from `eo-version.txt`, gets rejected with a misleading "not available in Maven Central" error and the CLI exits, whenever Maven Central returns a transient 5xx/429 instead of throwing a network exception.

## What Should Happen

`exists()` should only report `false` (confirmed absent) on an actual `404`. Any other non-200 status (5xx, 429, etc.) should be treated the same as a network exception: warn and assume the version exists, since it does not confirm the version is absent.

Thayorns:
`parser-version.js`'s `exists()` reduced any response to `res.statusCode === 200`, so every non-200 status collapsed to `false`. A transient `503`/`502`/`429` from Maven Central was therefore indistinguishable from a real `404`, and the CLI rejected a perfectly valid parser version — even the bundled default from `eo-version.txt` — with a misleading "not available in Maven Central" error. #998/PR #1020 fixed this for the exception path only; the response path stayed broken.

`exists()` now treats only a `404` as confirmed-absent. Any other non-200 status is handled the same way as a thrown network error: it warns (`Cannot verify parser version X in Maven Central (HTTP 503), proceeding anyway`) and assumes the version exists, since a non-404 does not confirm absence. A clean `200` still verifies silently.

The structure has one special case (`404` → absent) and a single general "otherwise it exists" branch, so it covers every other status rather than enumerating specific codes. Tests exercise the whole range: `503`, `429`, `502`, `500`, `301`, `403` all remain `true`, while `200`/`404` keep their existing behaviour, plus a check that the correct POM URL is still queried on an error response.

Verified: `npx mocha test/test_parser_version.js` 23/23 passing, `npx eslint .` clean.

Closes #1085.

### DIFF
--- a/src/parser-version.js
+++ b/src/parser-version.js
@@ -48,7 +48,16 @@ const request = require('sync-request'),
       if (ver && ver !== 'undefined') {
         try {
           const res = fetch('GET', version.url(ver), {timeout: 10000, socketTimeout: 10000});
-          result = res.statusCode === 200;
+          if (res.statusCode === 404) {
+            result = false;
+          } else {
+            if (res.statusCode !== 200) {
+              console.warn(colors.yellow(
+                `Cannot verify parser version ${ver} in Maven Central (HTTP ${res.statusCode}), proceeding anyway`
+              ));
+            }
+            result = true;
+          }
         } catch (e) {
           console.warn(colors.yellow(
             `Cannot verify parser version ${ver} in Maven Central (${e.message}), proceeding anyway`
--- a/test/test_parser_version.js
+++ b/test/test_parser_version.js
@@ -90,6 +90,42 @@ describe('parser-version', () => {
       const exists = parserVersion.exists('0.28.11', () => ({statusCode: 404}));
       assert.strictEqual(exists, false, 'a confirmed 404 must report the version as absent');
     });
+    it('assumes the version exists when Maven Central returns a transient 503', () => {
+      const exists = parserVersion.exists('0.28.11', () => ({statusCode: 503}));
+      assert.strictEqual(exists, true, 'a transient 503 must not be reported as a missing version');
+    });
+    it('assumes the version exists when Maven Central returns 429 rate limited', () => {
+      const exists = parserVersion.exists('0.28.11', () => ({statusCode: 429}));
+      assert.strictEqual(exists, true, 'a 429 rate limit must not be reported as a missing version');
+    });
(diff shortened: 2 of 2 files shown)


Thayorns:
@yegor256 take a look please.

Thayorns:
@yegor256 have a look

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+249](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1095, opened 2026-07-24
fix(#1089): preserve the original XSL error as cause when docs wraps it

### LINKED ISSUE #1089: `docs.js` doesn't preserve original XSL error's cause when wrapping it
## What Happens

`src/commands/docs.js` (lines 72-74), in the `createXmirHtmlBlock` helper:

```js
} catch(error) {
  throw new Error(`Error while applying XSL to XMIR: ${error.message}`, error);
}
```

The second argument to the `Error` constructor is supposed to be an options object like `{cause: error}`. Passing the raw `error` object directly does nothing: `'cause' in error` is false, so `.cause` is never set on the thrown error. This silently discards the original SaxonJS/XSLT error and its stack trace.

Every other place in this codebase that wraps an error uses the correct pattern, for example `src/commands/normalize.js:31` and `src/jdk.js:63`: `new Error(msg, {cause: e})`. `docs.js` is the inconsistent outlier.

Because Node's default uncaught-exception printer automatically prints a nested `[cause]` stack when `.cause` is set, this bug also means a real user loses the actual XSLT failure location when `docs` fails, seeing only the wrapper message with no indication of where in the XSL transform things went wrong.

Verified with a direct comparison of both forms:

```
new Error(msg, error)          -> outer.cause === undefined
new Error(msg, {cause: error}) -> outer.cause is the original Error, and Node prints its full stack under "[cause]:" when uncaught
```

This is a separate bug from the already-open `constructor` prototype-pollution crash (#1083) and from the already-known "dumps stack trace before rethrowing" wart in `docs.js`'s outer catch block (lines 180-183). This one is specifically about the `createXmirHtmlBlock` helper silently losing the cause on line 73.

## What Should Happen

Line 73 should be `throw new Error(\`Error while applying XSL to XMIR: ${error.message}\`, {cause: error});` so the original error and its stack are preserved, matching the pattern already used elsewhere in this codebase.

Thayorns:
`createXmirHtmlBlock` in `src/commands/docs.js` wrapped a failure with `new Error(msg, error)`. The second `Error` argument must be an options object; a raw error is ignored, so `.cause` was never set and the original SaxonJS/XSLT error and its stack were silently discarded. A user whose `docs` run fails then sees only the wrapper message with no indication of where in the transform it went wrong.

Passing `{cause: error}` fixes it, matching the pattern already used in `normalize.js` and `jdk.js`, and lets Node print the nested `[cause]` stack when the failure is uncaught.

Tests cover three angles: the helper keeps the original error as `.cause`, that cause survives to the top of the `docs` command through the outer catch, and the wrapper message still embeds the original error text.

Verified: `npx mocha test/commands/test_docs.js` 12/12 passing, `npx eslint .` clean.

Closes #1089.

### DIFF
--- a/src/commands/docs.js
+++ b/src/commands/docs.js
@@ -70,7 +70,7 @@ function createXmirHtmlBlock(filepath) {
     const xsl = fs.readFileSync(path.join(__dirname, '..', 'resources', 'xmir-transformer.xsl')).toString();
     return convertMarkdownToHtml(transformDocument(xmir, xsl));
   } catch(error) {
-    throw new Error(`Error while applying XSL to XMIR: ${error.message}`, error);
+    throw new Error(`Error while applying XSL to XMIR: ${error.message}`, {cause: error});
   }
 }
 
@@ -183,3 +183,5 @@ module.exports = function(opts) {
     }
   });
 };
+
+module.exports.createXmirHtmlBlock = createXmirHtmlBlock;
--- a/test/commands/test_docs.js
+++ b/test/commands/test_docs.js
@@ -253,4 +253,58 @@ describe('docs', () => {
     assert(test_content.includes('<code>Code test</code>'), `Markdown not processed correctly in ${test_html}`);
     done();
   });
+  /**
+   * Tests that createXmirHtmlBlock keeps the original error as the cause
+   * when the XSL transform fails, instead of silently discarding it.
+   * @param {Mocha.Done} done - Mocha callback signaling asynchronous completion
+   */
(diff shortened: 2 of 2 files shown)


Thayorns:
@yegor256 take a look please.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+265](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1099, opened 2026-07-24
Fail fast in java dataize when the JDK is missing

### LINKED ISSUE #1087: `dataize.js` crashes without informative error when JDK is missing
## What Happens

Every other Java-related command (`compile.js`, `link.js`, `test.js`, `pipeline.js`) calls `verifyJavac()` first, which gives a friendly, actionable error ("`javac` was not found on your PATH... Please install OpenJDK...") if the JDK is missing.

`src/commands/java/dataize.js` (line 26) does neither: it does not call `verifyJavac()`, and it only listens for the `'close'` event on the spawned child process, not `'error'`:

```js
spawn('java', params, {stdio: 'inherit'}).on('close', (code) => { ... });
```

When `java` isn't on PATH, Node emits an `'error'` event with no listener attached, which Node treats as an unhandled event and crashes the whole process with an internal stack trace, not a friendly message.

Verified by running directly in an environment with no `java` on PATH:

```
$ node -e "require('.../src/commands/java/dataize.js')('main.foo', [], {target:'.', stack:'64M', heap:'256M'})"
node:events:487
      throw er; // Unhandled 'error' event
Error: spawn java ENOENT
Emitted 'error' event on ChildProcess instance at: ...
```

## What Should Happen

`dataize` should either call `verifyJavac()` up front like its sibling commands, or at minimum attach an `'error'` handler on the spawned process and report a clear message (for example "java executable not found, please install a JDK").

Thayorns:
The `dataize` command was the only Java command that spawned a JVM without first checking that a JDK is present. When `java` was missing from the PATH, the spawned child emitted an `'error'` event with no listener, so Node crashed the whole process with an internal stack trace instead of a readable message.

This makes `dataize` verify the JDK up front with `verifyJavac()`, exactly like its siblings `compile`, `link`, `test`, and `pipeline`. A missing or broken toolchain now fails fast with the same actionable "install a JDK" guidance the rest of the toolchain already gives, before any subprocess is spawned. The command runner is injectable so the behaviour can be exercised without touching a real environment.

Tests cover three failure modes of the check: `javac` absent from the PATH, `javac` exiting non-zero, and the underlying reason being surfaced.

Closes #1087

### DIFF
--- a/src/commands/java/dataize.js
+++ b/src/commands/java/dataize.js
@@ -5,14 +5,17 @@
 
 const path = require('path');
 const {spawn} = require('node:child_process');
+const {verifyJavac} = require('../../jdk');
 
 /**
  * Runs the single executable binary.
  * @param {String} obj - Name of object to dataize
  * @param {Array} args - Arguments
  * @param {Object} opts - All options
+ * @param {Function} [exec] - Optional command runner for the JDK check
  */
-module.exports = function(obj, args, opts) {
+module.exports = function(obj, args, opts, exec) {
+  verifyJavac(exec);
   const params = [
     '-Dfile.encoding=UTF-8',
     `-Xss${opts.stack}`,
--- a/test/commands/test_dataize.js
+++ b/test/commands/test_dataize.js
@@ -6,6 +6,7 @@
 const assert = require('assert');
 const fs = require('fs');
 const path = require('path');
+const dataize = require('../../src/commands/java/dataize');
 const {runSync, parserVersion, homeTag, weAreOnline} = require('../helpers'),
 
   options = [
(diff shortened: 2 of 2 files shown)


Thayorns:
@yegor256 this makes `dataize` fail fast with the same JDK guidance as the other Java commands — happy to clarify or adjust anything if needed.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+233](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1104, opened 2026-07-25
Closes #1103. Check pinned version before audit

### LINKED ISSUE #1103: audit command ignores the --pin version check
## Problem

The global `--pin <version>` option is documented as:

> Fail if eoc version doesn't match exactly

However, the `audit` command ignores this option and continues executing
even when the requested version does not match the current EOC version.

The `audit` action currently calls the command directly:

```js
program.command('audit')
  .description('Inspect all packages and report their status')
  .action(async (str, opts) => {
    await coms().audit(program.opts());
  });
```

Unlike most other commands, it does not call:

```js
pin(program.opts());
```

before executing.

## Steps to reproduce

Run from the repository root:

```bash
node src/eoc.js --pin=999.999.999 audit
echo "Exit code: $?"
```

## Actual behavior

The `audit` command executes normally, prints the Maven version, and exits
successfully:

```text
Apache Maven 3.9.16
...
Exit code: 0
```

## Expected behavior

The command should reject the mismatched pinned version before running
the audit operation:

```text
Version mismatch: you are running eoc 0.0.0, but --pin option requires 999.999.999
```

It should exit with a non-zero status code.

## Suggested solution

Call `pin(program.opts())` in the `audit` command action before invoking
`coms().audit(...)`:

```js
program.command('audit')
  .description('Inspect all packages and report their status')
  .action(async (str, opts) => {
    pin(program.opts());
    await coms().audit(program.opts());
  });
```

A regression test should verify that `audit` fails when `--pin` does not
match the current EOC version.

theaflit:
Closes #1103.

The `audit` command ignored the global `--pin` option and continued
executing when the pinned version did not match the current EOC version.

This change checks the pinned version before running `audit` and adds
a regression test for the mismatched version case.

Tests:
- `npx mocha test/test_eoc.js --timeout 1200000` — 28 passing
- `npm test` — 150 passing, 10 pending
- `npx grunt` — 150 passing, 10 pending

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -166,6 +166,7 @@ program.hook('preAction', (command) => {
 program.command('audit')
   .description('Inspect all packages and report their status')
   .action(async (str, opts) => {
+    pin(program.opts());
     await coms().audit(program.opts());
   });
 
--- a/test/test_eoc.js
+++ b/test/test_eoc.js
@@ -250,6 +250,13 @@ describe('eoc', () => {
 
 describe('eoc', () => {
   before(weAreOnline);
+  it('fails audit due to version mismatch if different --pin provided', (done) => {
+    assert.throws(
+      () => { runSync(['--pin=29.9.4', 'audit']); },
+      /Version mismatch: you are running eoc [0-9]+\.[0-9]+\.[0-9]+, but --pin option requires 29.9.4/
+    );
+    done();
+  });
   it('cleans successfully when if --pin not provided', (done) => {
     const stdout = runSync(['clean']);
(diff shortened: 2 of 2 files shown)


theaflit:
@yegor256 Could you please review this PR when you have time?

0crat:
@theaflit Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+30](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1107, opened 2026-07-25
Closes #1106. Guard clean against symbolic links

### LINKED ISSUE #1106: clean can delete the current working directory through a symbolic link
## Problem

The `clean` command is intended to refuse deletion of the current working
directory, its ancestors, and the user's home directory.

However, this protection can be bypassed when `--target` reaches the current
working directory through a symbolic link.

The guard compares lexical paths using `path.relative()`, but does not resolve
symbolic links before performing the safety check.

## Steps to reproduce

Run from the repository root:

```bash
EOC="$PWD/src/eoc.js"
ROOT="$(mktemp -d)"

mkdir -p "$ROOT/real"
printf "do not delete\n" > "$ROOT/real/marker.txt"

ln -s "$ROOT" "$ROOT/link"

echo "Temporary root: $ROOT"
echo "Before:"
ls -la "$ROOT/real"

(
  cd "$ROOT/real"
  node "$EOC" --target "$ROOT/link/real" clean
)

echo "Exit code: $?"

if [ -d "$ROOT/real" ]; then
  echo "SAFE: real directory still exists"
else
  echo "BUG: clean deleted its current working directory through a symlink"
fi

rm -rf "$ROOT"
```

## Actual behavior

```text
The directory ../link/real was deleted
Exit code: 0
BUG: clean deleted its current working directory through a symlink
```

The target path is a symbolic-link alias of the current working directory, but
the command deletes it successfully.

## Expected behavior

The command should detect that the resolved target refers to the current
working directory and refuse to delete it with a non-zero exit code.

## Suggested solution

Resolve existing target paths to their canonical filesystem locations before
performing the safety comparison.

A regression test should verify that `clean` refuses to delete the current
working directory or one of its ancestors when the target reaches it through
a symbolic link.

theaflit:
Closes #1106.

The `clean` command compared lexical paths when checking whether the
target was the current working directory, one of its ancestors, or the
home directory. This allowed the protection to be bypassed through a
symbolic link.

This change resolves symbolic links in the target's parent path before
performing the safety comparison. The original target remains unchanged
for the actual deletion.

A regression test verifies that `clean` refuses to delete the current
working directory through a symbolic link.

Tests:
- `npx mocha test/commands/test_clean.js --timeout 1200000` — 6 passing
- `npm test` — 150 passing, 10 pending
- `npx grunt` — 150 passing, 10 pending

### DIFF
--- a/src/commands/clean.js
+++ b/src/commands/clean.js
@@ -9,18 +9,32 @@ const path = require('path');
 const os = require('os');
 
 /**
- * Refuses, by throwing, to delete a directory that is the current
+ * Resolves symbolic links in the parent part of a path.
+ * @param {String} target - Path to canonicalize
+ * @return {String} Path with its parent resolved
+ */
+function canonical(target) {
+  const parent = path.dirname(target);
+  return fs.existsSync(parent)
+    ? path.join(fs.realpathSync(parent), path.basename(target))
+    : target;
+}
+
+/**
+ * Refuses, by throwing, to delete a directory that resolves to the current
  * working directory, an ancestor of it, or the user's home directory;
- * otherwise returns it unchanged, so the guard cannot be skipped or
- * reordered away from the value it protects.
+ * otherwise returns the original target unchanged, so the guard cannot be
+ * skipped or reordered away from the value it protects.
  * @param {String} target - Resolved absolute path of the directory to delete
- * @return {String} The same target, once confirmed safe to delete
+ * @return {String} The original target, once confirmed safe to delete
  */
 function guarded(target) {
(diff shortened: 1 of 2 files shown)


0crat:
@theaflit Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+46](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1112, opened 2026-07-27
Validate --object format before parsing to avoid crash

### LINKED ISSUE #1086: `eoc test --object` crashes when input lacks two dot-separated segments

Thayorns:
`eoc test --object <name>` split the value on dots and popped two segments off the end without checking there were two. A single-segment value like `--object=app`, or any typo missing a dot, left the array empty and `.replace()` was called on `undefined`, crashing the process with a raw `TypeError` stack trace instead of telling the user what was wrong.

This adds a fail-fast guard: if the value has fewer than two dot-separated segments, or any segment is empty (`app.`, `.works-fine`, `..`), it throws a clear message — `Invalid --object format: "...", expected object.method (or pkg.object.method)` — before any parsing happens. Valid names, including deep packages like `foo.bar.app.works-fine`, are unaffected.

Covered by tests for each rejected shape plus a deep-package happy path.

Closes #1086

### DIFF
--- a/src/commands/java/test.js
+++ b/src/commands/java/test.js
@@ -22,6 +22,11 @@ module.exports = function(opts, maven = mvnw) {
   ];
   if (opts.object) {
     const parts = opts.object.split('.');
+    if (parts.length < 2 || parts.some((part) => part === '')) {
+      throw new Error(
+        `Invalid --object format: "${opts.object}", expected object.method (or pkg.object.method)`
+      );
+    }
     const method = parts.pop().replace(/-/g, '_');
     const obj = parts.pop().replace(/-/g, '_');
     const pkg = parts.map((p) => `EO${p.replace(/-/g, '_')}`).join('.');
--- a/src/commands/java/test.js
+++ b/src/commands/java/test.js
@@ -22,16 +22,15 @@ module.exports = function(opts, maven = mvnw) {
   ];
   if (opts.object) {
     const parts = opts.object.split('.');
-    if (parts.length < 2 || parts.some((part) => part === '')) {
+    if (parts.some((part) => part === '')) {
       throw new Error(
-        `Invalid --object format: "${opts.object}", expected object.method (or pkg.object.method)`
+        `Invalid --object format: "${opts.object}", expected object, object.method or pkg.object.method`
       );
     }
-    const method = parts.pop().replace(/-/g, '_');
-    const obj = parts.pop().replace(/-/g, '_');
+    const method = parts.length > 1 ? `#${parts.pop().replace(/-/g, '_')}` : '';
+    const cls = `EO${parts.pop().replace(/-/g, '_')}*Test`;
     const pkg = parts.map((p) => `EO${p.replace(/-/g, '_')}`).join('.');
-    const cls = `EO${obj}*Test`;
-    args.push(`-Dtest=${pkg ? `org.eolang.${pkg}.${cls}` : `org.eolang.${cls}`}#${method}`);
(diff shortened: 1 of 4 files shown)


Thayorns:
@yegor256 take a look

yegor256:
@Thayorns this is not correct. Consider this code:

```
[] > app
  42 > @
```

I should be able to run this:

```
$ eoc dataize app
```

And I should get `42`.

Thayorns:
@yegor256 you're right that `app` is a valid object, and I've changed the code accordingly — but I want to flag that `eoc dataize app` was never affected by this PR, in case that's what prompted the comment.

`dataize` takes its object as a positional argument and passes it straight through:

```js
// src/eoc.js:306
await coms().dataize(program.args[1], program.args.slice(2), {...program.opts(), ...str});
```

The validation this PR added lives in `src/commands/java/test.js` and only ever read `opts.object`, which is set by the `--object` option of the `test` command — the sole occurrence of `object` in `src/eoc.js`. So your example worked before this PR and still does. I've added a test running exactly it through the full pipeline, and it passes:

```
[0x40450000-00000000-] = 42.0, or "..."
Finished in 0.03s (8 bytes)
```

Where your point does land is the premise behind the guard. `app` is a complete object, so treating a single-segment `--object` as malformed refused a meaningful name. That's now fixed — a bare object name runs every test in it:

```
app                    -> -Dtest=org.eolang.EOapp*Test
app.works-fine         -> -Dtest=org.eolang.EOapp*Test#works_fine
foo.bar.app.works-fine -> -Dtest=org.eolang.EOfoo.EObar.EOapp*Test#works_fine
```

Multi-segment values keep their existing meaning, so `foo.app` stays object plus method — reading it as package plus object would be ambiguous and would change what `app.works-fine` means. Only empty segments (`app.`, `.works-fine`, `..`) are still rejected, which is the case that originally crashed on `.replace()` of `undefined`.

Worth noting this now contradicts the "What Should Happen" in #1086, which asked for single-segment values to be rejected with a clear message. I've taken your call over the issue text.

yegor256:
@Thayorns my mistake, sorry

Thayorns:
@yegor256 reverted changes, have a look

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+271](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1129, opened 2026-07-29
chore: bump fast-xml-parser to 5.10.1 / 4.5.7 (dependency hygiene)

anupamme:
## Summary

Routine dependency hygiene bump for `fast-xml-parser`:

- Root dependency: `5.9.3` → `5.10.1` (clears `GHSA-8r6m-32jq-jx6q`, which affects `>= 5.9.3, < 5.10.1`)
- `eo2js` nested dependency: `4.5.3` → `4.5.7` (via `overrides` block — see below)

## What changed

**`package.json`**
- `fast-xml-parser` range: `^5.2.3` → `^5.10.1`
- Added `overrides.eo2js.fast-xml-parser: "^4.5.4"` to force `eo2js`'s transitive dep off `4.5.3`

**`package-lock.json`**
- Regenerated via `npm install` (not hand-edited)

## Risk assessment

This is a Node CLI. There is no browser, DOM, or rendered HTML. The advisories patched here are entity-expansion DoS variants (not XSS). Neither XML input in this project is attacker-controlled:

- `src/parser-version.js:24` parses `maven-metadata.xml` fetched over HTTPS from `repo.maven.apache.org`
- `eo2js` parses XMIR from the local build output of the EO compiler

Realistic worst case is your own build hanging. The bump is worth having to keep `npm audit` quiet and the dependency tree current.

## `overrides` block

The override is a temporary workaround until `eo2js` updates its own range. Tracked in #1130 for removal.

## Surviving audit finding

`npm audit` still reports `GHSA-gh4j-gqv2-49f6` (medium, XMLBuilder) against the nested `4.5.7`. No 4.x patch exists. Acceptable: `eo2js` imports only `XMLParser`, never `XMLBuilder` — the vulnerable code path is unreachable.

## Verification

- [x] `npm ci` succeeds
- [x] `npm ls fast-xml-parser` resolves to `5.10.1` (root) and `4.5.7` (`eo2js` nested)
- [x] Build passes (`npx grunt`)
- [x] Lint passes (`npx eslint`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "colors": "^1.4.0",
         "commander": "^12.1.0",
         "eo2js": "^0.0.8",
-        "fast-xml-parser": "^5.2.3",
+        "fast-xml-parser": "^5.3.5",
         "marked": "^4.3.0",
         "relative": "^3.0.2",
         "semver": "^7.7.2",
@@ -875,18 +875,6 @@
       "peerDependencies": {
         "@langchain/core": "^1.2.2"
       }
-    },
-    "node_modules/@nodable/entities": {
-      "version": "2.2.0",
-      "resolved": "https://registry.npmjs.org/@nodable/entities/-/entities-2.2.0.tgz",
-      "integrity": "sha512-9uGyhaQavEUMC8AIddIjau4NsnsXhou+j5sBAGojCM1oxmQpVKTWR/9JxABD6UAv12vpIms55fPZKFQEhG6uBg==",
-      "funding": [
-        {
-          "type": "github",
-          "url": "https://github.com/sponsors/nodable"
-        }
-      ],
-      "license": "MIT"
     },
     "node_modules/@pkgjs/parseargs": {
       "version": "0.11.0",
@@ -2219,26 +2207,10 @@
       "dev": true,
       "license": "MIT"
     },
-    "node_modules/fast-xml-builder": {
-      "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/fast-xml-builder/-/fast-xml-builder-1.2.0.tgz",
-      "integrity": "sha512-00aAWieqff+ZJhsXA4g1g7M8k+7AYoMUUHF+/zFb5U6Uv/P0Vl4QZo84/IcufzYalLuEj9928bXN9PbbFzMF0Q==",
-      "funding": [
-        {
-          "type": "github",
-          "url": "https://github.com/sponsors/NaturalIntelligence"
-        }
-      ],
-      "license": "MIT",
-      "dependencies": {
-        "path-expression-matcher": "^1.5.0",
-        "xml-naming": "^0.1.0"
-      }
-    },
     "node_modules/fast-xml-parser": {
-      "version": "5.9.3",
-      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-5.9.3.tgz",
-      "integrity": "sha512-brCNCeScma/kqa54J4PIDriSSSLssRkuYaUCpvHJulGc3HGI/xxKUCTDcYkAdqJsyb//ydpbxecjC3hB9+tb/g==",
+      "version": "5.3.5",
+      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-5.3.5.tgz",
+      "integrity": "sha512-JeaA2Vm9ffQKp9VjvfzObuMCjUYAp5WDYhRYL5LrBPY/jUDlUtOvDfot0vKSkB9tuX885BDHjtw4fZadD95wnA==",
       "funding": [
         {
           "type": "github",
@@ -2247,12 +2219,7 @@
       ],
       "license": "MIT",
       "dependencies": {
-        "@nodable/entities": "^2.2.0",
-        "fast-xml-builder": "^1.2.0",
-        "is-unsafe": "^1.0.1",
-        "path-expression-matcher": "^1.5.0",
-        "strnum": "^2.4.1",
-        "xml-naming": "^0.1.0"
+        "strnum": "^2.1.2"
       },
       "bin": {
         "fxparser": "src/cli/cli.js"
@@ -3369,18 +3336,6 @@
       "funding": {
         "url": "https://github.com/sponsors/sindresorhus"
       }
-    },
-    "node_modules/is-unsafe": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/is-unsafe/-/is-unsafe-1.0.1.tgz",
-      "integrity": "sha512-CLK2+VdgERgD96EYm5lUQssZYlRg2tkZnbsxZoacmSiRxiFJ4Nk4SzjCl+Ur+v3kXIY9dTIdb3IH22y1mZ56LA==",
-      "funding": [
-        {
-          "type": "github",
-          "url": "https://github.com/sponsors/NaturalIntelligence"
-        }
-      ],
-      "license": "MIT"
     },
     "node_modules/is-windows": {
       "version": "1.0.2",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -2219,26 +2207,10 @@
...
-      "dependencies": {
-        "path-expression-matcher": "^1.5.0",
-        "xml-naming": "^0.1.0"
-      }
-    },
     "node_modules/fast-xml-parser": {
-      "version": "5.9.3",
-      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-5.9.3.tgz",
-      "integrity": "sha512-brCNCeScma/kqa54J4PIDriSSSLssRkuYaUCpvHJulGc3HGI/xxKUCTDcYkAdqJsyb//ydpbxecjC3hB9+tb/g==",
+      "version": "5.3.5",
```
Thayorns: This is a downgrade, not an upgrade — master resolves 5.9.3 here, and this pins it back to 5.3.5.

5.9.3 is already past the CVE-2026-25896 fix (patched in 5.3.5), so this buys nothing and reintroduces everything patched in between:

| Advisory | Severity | Patched in |
|---|---|---|
| CVE-2026-26278 — DoS via DOCTYPE entity expansion | high | 5.3.6 |
| CVE-2026-27942 — stack overflow in XMLBuilder | low | 5.3.8 |
| CVE-2026-33036 — numeric entity expansion bypass | high | 5.5.6 |
| CVE-2026-33349 — entity limits bypassed when set to zero | medium | 5.5.7 |
| CVE-2026-41650 — comment/CDATA injection in XMLBuilder | medium | 5.7.0 |

Net effect: two highs and two mediums added, in exchange for a finding the tree didn't have. A re-scan against 5.3.5 should have surfaced these — worth checking why "Scanner re-scan confirms fix" passed.

Please regenerate the lockfile with `npm install` against `^5.10.1` instead of editing it by hand.

package.json
```diff
@@ -9,7 +9,7 @@
     "colors": "^1.4.0",
     "commander": "^12.1.0",
     "eo2js": "^0.0.8",
-    "fast-xml-parser": "^5.2.3",
+    "fast-xml-parser": "^5.3.5",
```
Thayorns: `^5.10.1` would be the right target here — 5.10.1 is the current latest and clears `GHSA-8r6m-32jq-jx6q` (`>= 5.9.3, < 5.10.1`), which master genuinely is affected by. Note this range bump alone changes nothing in practice — the lockfile is what installs — so it needs the regenerated lockfile alongside it.

Thayorns (changes requested):
Thanks for the scan, but I think this change goes the wrong direction and misses the copy that's actually vulnerable.

**The top-level dependency was never affected.** `GHSA-m7jm-9gc2-mpf2` (CVE-2026-25896) covers `>= 5.0.0, < 5.3.5`. Master's lockfile resolves **5.9.3**, which is outside that range. The scanner most likely matched the declared range `^5.2.3` in `package.json` rather than the resolved version.

**The vulnerable copy is the nested one, and it isn't touched.** `node_modules/eo2js/node_modules/fast-xml-parser` is pinned at **4.5.3**, and CVE-2026-25896 affects `>= 4.1.3, < 4.5.4`. That's the genuine hit here — presumably where the "4.5.4" in the title came from — but this PR leaves it at 4.5.3. Fixing it means getting `eo2js` to relax its `^4.3.5` range, or adding an `overrides` entry.

**Master is affected by a different advisory this PR doesn't mention.** `GHSA-8r6m-32jq-jx6q` (high — repeated DOCTYPE declarations reset entity expansion limits) covers `>= 5.9.3, < 5.10.1`, so 5.9.3 does need a bump — forward to **5.10.1**, the current latest.

On scope: the only consumer is `src/parser-version.js:24`, which parses `maven-metadata.xml` fetched over HTTPS from `repo.maven.apache.org`. Nothing user-supplied reaches `XMLParser`, so the "handles user-influenced input" / "likely exploitable" framing in the description doesn't hold for this codebase. Worth fixing regardless, just not at CRITICAL urgency.

Suggested replacement: bump the top-level to `^5.10.1` with a lockfile regenerated by `npm install` (not hand-edited), and add an `overrides` entry to pull the nested `eo2js` copy up to `4.5.4`+.

anupamme:
Code review comments addressed. pls review.

package-lock.json
anupamme: addressed.

package.json
anupamme: addressed.

package.json
```diff
@@ -73,5 +73,10 @@
     "coverage": "nyc --reporter=lcov --reporter=text-summary mocha test/**/*.js --timeout 1200000",
     "test": "mocha 'test/**/*.js' --timeout 1200000"
   },
+  "overrides": {
+    "eo2js": {
+      "fast-xml-parser": "^4.5.4"
+    }
+  },
```
Thayorns, line 80: This block is the right mechanism and it works — `npm ls` confirms `eo2js` resolves to 4.5.7. It's also the piece with ongoing cost: it permanently pins a transitive dep, and someone has to remember to drop it once `eo2js` relaxes its own `^4.3.5` range. Worth a linked issue to remove it later, since `package.json` can't carry a `@todo` puzzle.

Given that cost, it's worth being accurate about what this PR is buying — because for this codebase it is close to nothing, security-wise.

**The XSS framing in the description doesn't apply here.** `eoc` is a Node CLI. There is no browser, no DOM, no rendered HTML in any of these paths. That wording is copied from the upstream advisory without checking whether it transfers, and it doesn't.

**Stripped of the labels, these CVEs are entity-expansion DoS (billion-laughs variants) plus a regex injection in DOCTYPE entity-name handling.** Realistic impact: the parsing process hangs or exhausts memory.

**Neither XML input in this project is attacker-controlled:**
- `src/parser-version.js:24` parses `maven-metadata.xml` fetched over HTTPS from `repo.maven.apache.org`. Feeding it hostile input means compromising Maven Central or breaking TLS — and at that point you control which `eo-maven-plugin` jar `eoc` downloads and executes (`src/mvnw.js:12`, `src/eoc.js:111`). The XML parser is nowhere near the weakest link in that scenario.
- `eo2js/src/commands/transpile.js:134` parses XMIR from the local build target dir, generated by the EO compiler from the user's own `.eo` sources. If you're running `eoc` over untrusted EO code, that code gets compiled and run — the parser isn't the trust boundary there either.

So the worst realistic outcome is your own build hanging, on a machine where an attacker already had a much better option.

**I'm not arguing to close this.** The bump is worth having on dependency-hygiene grounds: the deps genuinely were stale, 4.5.3 was six patch releases behind, and a permanently-red `npm audit` is noise that will bury a real finding later. But please retitle and rewrite the body to describe it as a routine dependency bump, and drop the CRITICAL / "likely exploitable" / "handles user-influenced input" claims — none of them hold for this project. The current title still reads "upgrade to 5.3.5, 4.5.4", which now contradicts the diff entirely.

That matters more than usual here: the first revision of this PR proposed a *downgrade* from 5.9.3 to 5.3.5 that would have reintroduced four advisories, and reported "Scanner re-scan confirms fix". Accurate framing is what lets a reviewer catch that.

FYI, one finding survives and is fine to ignore: `npm audit` still reports GHSA-gh4j-gqv2-49f6 (medium, XMLBuilder comment/CDATA injection) against the nested 4.5.7 — its range is `< 5.7.0` with no 4.x patch ever published. It's unreachable, since `eo2js` imports only `XMLParser`, never `XMLBuilder`.

Thayorns (commented):
The code in this revision is correct — I verified it independently rather than going by the re-scan: `npm ci` succeeds against the lockfile, the tree resolves to `fast-xml-parser@5.10.1` top-level and `eo2js → fast-xml-parser@4.5.7 overridden`, and both copies still parse a `maven-metadata.xml` shape correctly. Both of my earlier points are addressed.

One comment inline on the `overrides` block about how this should be framed before it merges.

@yegor256 take a look at this one

package.json
anupamme: Agreed on all points — thanks for the thorough breakdown.

Done:
- PR retitled and description rewritten to frame this as routine dependency hygiene (no XSS/CRITICAL/exploitable language)
- Created #1130 to track removal of the `overrides` block once `eo2js` updates its range
- Noted the surviving `GHSA-gh4j-gqv2-49f6` finding and why it's acceptable (XMLBuilder unreachable)

0crat:
@anupamme Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+392](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1131, opened 2026-07-30
fix(#1120): use maximum heap limit for Java runner

### LINKED ISSUE #1120: `--heap` sets the JVM's initial heap (`-Xms`) instead of its maximum (`-Xmx`)
## What happens

`eoc dataize`'s Java runner (`src/commands/java/dataize.js:22`) passes the `--heap` option to the JVM as `-Xms`, the *initial* heap size, not the *maximum*:

```js
`-Xms${opts.heap}`,
```

`--heap <size>` is documented in `src/eoc.js:300` as "Set the heap size for the VM" with default `256M`, and its evident purpose (matching `--stack`/`-Xss` right above it, which is correctly wired) is to bound memory usage and prevent `OutOfMemoryError`. Because it is wired to `-Xms` instead of `-Xmx`, it does not do that: the JVM's actual heap ceiling stays at whatever the JVM's own ergonomics pick, regardless of `--heap`.

Verified on JDK 21:
- `java -Xms256M -XX:+PrintFlagsFinal -version | grep MaxHeapSize` → `MaxHeapSize = 4018143232` (the JVM default ceiling, ignoring the 256M initial size).
- `java -Xmx6G -XX:+PrintFlagsFinal -version | grep MaxHeapSize` → `MaxHeapSize = 6442450944` (correctly reflects the requested value).
- `java -Xms64G -version` → JVM startup aborts: `Error occurred during initialization of VM / insufficient memory for the Java Runtime Environment to continue`. `java -Xmx64G -version` starts normally.

So a user passing `--heap 6G` to raise the ceiling gets no effect (heap ceiling stays at the default), and a user passing a large `--heap` value to be generous can crash the JVM outright at startup, both because the flag is bound to the wrong JVM option.

## What should happen

`` `-Xms${opts.heap}` `` should be `` `-Xmx${opts.heap}` ``, so `--heap` actually bounds/raises the JVM's maximum heap as documented.

## Steps to reproduce

1. Run `eoc dataize <obj> --heap=6G` on a project.
2. Inspect the JVM's actual `MaxHeapSize` (e.g. via `-XX:+PrintFlagsFinal`), or reproduce an `OutOfMemoryError` that `--heap=6G` was supposed to prevent.
3. Observe the ceiling is unaffected by the `--heap` value.

webhop123:
Closes #1120

## What changed

The Java `dataize` runner now maps `--heap` to `-Xmx` instead of `-Xms`,
so the option controls the JVM maximum heap as documented.

The Java process runner is injectable for unit testing. A regression test
captures the generated arguments and verifies that `-Xmx256M` is present
and no `-Xms` argument remains.

## Verification

- `npx mocha test/commands/test_dataize.js --grep "dataize/java"`
  - 4 tests passing
- `npx eslint .`
  - passed
- `npx grunt --file=test/commands/test_dataize.js`
  - the 4 unit tests passed
  - the 2 CLI integration cases could not run locally on Windows because
    their pipeline did not create `target/eoc.jar`

@yegor256

### DIFF
--- a/src/commands/java/dataize.js
+++ b/src/commands/java/dataize.js
@@ -13,20 +13,21 @@ const {verifyJavac} = require('../../jdk');
  * @param {Array} args - Arguments
  * @param {Object} opts - All options
  * @param {Function} [exec] - Optional command runner for the JDK check
+ * @param {Function} [runner] - Optional Java process runner
  */
-module.exports = function(obj, args, opts, exec) {
+module.exports = function(obj, args, opts, exec, runner = spawn) {
   verifyJavac(exec);
   const params = [
     '-Dfile.encoding=UTF-8',
     `-Xss${opts.stack}`,
-    `-Xms${opts.heap}`,
+    `-Xmx${opts.heap}`,
     '-jar', path.resolve(opts.target, 'eoc.jar'),
     opts.verbose ? '--verbose' : '',
     obj,
     ...args,
   ].filter((i) => i);
   console.debug(`+ java ${params.join(' ')}`);
-  spawn('java', params, {stdio: 'inherit'}).on('close', (code) => {
+  runner('java', params, {stdio: 'inherit'}).on('close', (code) => {
     if (code !== 0) {
       console.error(`JVM failed with exit code ${code}`);
       process.exit(1);
--- a/test/commands/test_dataize.js
+++ b/test/commands/test_dataize.js
@@ -86,6 +86,27 @@ describe('dataize', () => {
 });
 
 describe('dataize/java', () => {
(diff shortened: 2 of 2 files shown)


0crat:
@webhop123 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1137, opened 2026-07-30
Add format command with --fix option

### LINKED ISSUE #1136: `format` command with `--fix` option is missing
Let's add `format` command with the `--fix` option

yegor256:
Adds an `eoc format` command. On its own it checks that every `.eo` source matches the canonical form the printer renders and exits non-zero when something diverges, so it can sit in CI. With `--fix` it passes `-Deo.autoFix=true` and the divergent files get rewritten in place. Both modes are just the `eo:format` goal of eo-maven-plugin, run right after `eo:register` in a single Maven invocation.

The old `fmt` command is gone. It did the reformatting by printing `1-parse` XMIR back over the sources, which worked but could only rewrite, never check, and it duplicated logic the plugin already has. Its test cases are ported over to `test/commands/test_format.js` and pass unchanged, plus three new ones for the check mode.

Two things worth knowing. The `eo:format` goal only exists since eo-maven-plugin 0.62.0, so this stacks on #1135 and on the test fixture updates that come with it — the two commits ahead of `master` here are exactly that. And in check mode the plugin prints a colored unified diff per file, but `mvnw` runs `--quiet` unless you pass `--verbose`, so by default you only see the summary line; happy to wire that up differently if you'd rather see the diff always.

Closes #1136


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (480 >= 100); -8 for way too many hits-of-code (480 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2780](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1142, opened 2026-07-31
Bump default EO version to 0.62.1

### LINKED ISSUE #1141: Default EO version 0.62.0 causes ClassNotFoundException and reference resolution issues
`eo-version.txt` and `home-tag.txt` both contain `0.62.0`. `src/eoc.js:113` reads the former as the default for `--parser` (declared on line 143), and `src/eoc.js:111` reads the latter as the default for `--home-tag` (line 142). A plain `eoc` invocation therefore compiles against EO 0.62.0 with no opt-in from the user.

EO 0.62.0 cannot run programs that use `malloc` or `tuple.each`. Two defects are open upstream: `Φ.chunk` is never probed, so `malloc.of` fails with `ClassNotFoundException: org.eolang.EOchunk` (objectionary/eo#6174), and bare same-package references inside `tuple/` resolve to `Φ.eachi` and `Φ.reducedi` instead of `Φ.tuple.*`, so `tuple.each` cannot resolve (objectionary/eo#6175). Since `malloc` underpins most of the object model, the default toolchain is unusable for anything non-trivial. Neither `objects/chunk.eo` nor `objects/tuple/eachi.eo` exists in objectionary/home at tag 0.61.3, so the previous default predates both defects — though I have not confirmed that 0.61.3 is otherwise green.

Nothing in eoc's own code is at fault; only the pinned default is, following the bump in #1134. Hold `eo-version.txt` and `home-tag.txt` at the last working EO release until objectionary/eo#6174 and objectionary/eo#6175 are fixed and released, then bump to the version that carries both fixes.

yegor256:
This moves `eo-version.txt` and `home-tag.txt` from 0.62.0 to 0.62.1 and adds a unit test that allocates a memory block through `malloc.of`.

The test is the point of the change. On 0.62.0 it fails with `Couldn't find object 'Φ.chunk' because there's no class 'org.eolang.EOchunk'`, which is objectionary/eo#6174 — `Φ.chunk` is never probed, so nothing pulls it in and `malloc.of` dies at runtime. Since malloc underpins most of the object model, the default toolchain was unusable for anything non-trivial. On 0.62.1 the test passes.

0.62.1 also carries the fix for objectionary/eo#6169, so `stdout` prints again. That let me drop the two `this.skip()` calls in the dataize tests and put the itest workflow back to dataizing `program` instead of the silent `story`. Both verified locally; the full suite is green.

Closes #1141


test/commands/test_test.js
```diff
@@ -73,6 +73,47 @@ describe('test', () => {
...
+        '        [m]',
+        '          seq > @',
+        '            *',
+        '              m.put 42',
+        '              m.get',
+        '      42',
+      ].join('\n')
+    );
+    assert.ok(
+      runSync([
```
Thayorns, line 101: This flag list is a copy of the one the `test` helper assembles at lines 37-51, and it has already drifted from it: the helper passes `--language=${lang}` on line 48, this copy passes nothing, so the test rides on the default declared at src/eoc.js:143. The next flag added to the helper will silently miss this test too.

test/commands/test_test.js
```diff
@@ -73,6 +73,47 @@ describe('test', () => {
...
+        '--easy',
+        '--blind',
+        `--parser=${parserVersion}`,
+        `--home-tag=${homeTag}`,
+        '--stack=16M',
+        '--heap=128M',
+        '-s', path.resolve(home, 'src'),
+        '-t', path.resolve(home, 'target'),
+      ]).includes('Tests run: 1, Failures: 0, Errors: 0, Skipped: 0'),
+      'malloc cannot be dataized by the pinned EO version'
```
Thayorns, line 113: When this assertion fails, the only thing a red build shows is this sentence — `runSync` is called inline, so the surefire output that would say what actually ran is discarded. `runSync` prints stdout only when the process exits non-zero (test/helpers.js:37), and the case this assertion guards is precisely maven exiting zero with a different count. The test right below on line 119 keeps stdout in a variable and appends it to the message.

Thayorns (commented):
Bumps the pinned EO version to 0.62.1, removes the two `this.skip()` guards in the dataize tests and puts the itest workflow back on `program`, and adds a unit test that dataizes `malloc.of`. The change is straightforward and the new test does pin down the behaviour the bump is meant to restore; two remarks below are about how that test is written, not about the bump itself.

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+432](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2892](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1150, opened 2026-08-01
fix(#1145): pass eo2js arguments as argv array instead of a shell string

### LINKED ISSUE #1145: Shell injection vulnerability and failure on paths with spaces in eoc -l js dataize/link/test/transpile
## Summary
src/eo2jsw.js builds the eo2js invocation by concatenating a plain string and running it with execSync, which spawns a shell:
```
execSync(
  `node ${bin} ${command} ${flags(args, lib).filter((flag) => flag !== '').join(' ')}`,
  {timeout: 1200000, windowsHide: true, stdio: 'inherit'}
);
```
src/commands/js/dataize.js feeds the object name straight into that string with no escaping:
```
return eo2jsw(
  ['dataize', obj, ...args.filter((arg) => !arg.startsWith('-'))].join(' '),
);
```
## Steps to reproduce
```
eoc -l js dataize "foo; touch /tmp/PWNED #"
```
## Actual behavior
The resulting shell command is:
```
node <bin> dataize foo; touch /tmp/PWNED # --target .eoc --project project ...
```
; terminates the dataize invocation and touch /tmp/PWNED runs as a second, injected shell command. Any object name that isn't a bare word (spaces, ;, `, $(), |, &&, quotes...) either breaks the CLI or executes attacker-controlled shell code — e.g. this is reachable through CI pipelines that pass object names derived from external input.

Separately (same root cause), any value with a space breaks argument parsing even without malicious intent, e.g. --target pointing at a Windows path with spaces:
```
--target /home/user/My Documents/.eoc
```
gets shell-split into --target /home/user/My plus a stray positional token Documents/.eoc, so eo2js receives the wrong target directory silently.

##  Expected behavior
Object names, paths and other values are passed to the child process as separate argv entries (no shell involved), exactly like src/commands/java/dataize.js already does via spawn(bin, params, {...}) with an array.

ImilB:
### Description

Replaces `execSync` with `execFileSync` in `eo2jsw.js` so arguments are passed as an explicit array directly to the Node process, completely bypassing shell interpretation. `js/dataize.js` now passes the sub-command and object name as separate array elements instead of joining them into a single space-separated string. Also splits the previously joined `'--foreign eo-foreign.json'` flag into two separate array entries, fixing the same root cause at the flag level.

Fixes #1145

### DIFF
--- a/src/eo2jsw.js
+++ b/src/eo2jsw.js
@@ -4,13 +4,13 @@
  */
 
 const path = require('path');
-const {execSync} = require('child_process'),
+const {execFileSync} = require('child_process'),
 
   /**
    * Convert eoc arguments to appropriate eo2js flags
    * @param {Object} args - eoc arguments
    * @param {String} lib - Path to eo2js lib
-   * @return {Array.<String>} - Flags for eo2js
+   * @return {Array.<String>} - Flags for eo2js, ready to pass as argv entries
    */
   flags = function(args, lib) {
     return [
@@ -20,28 +20,31 @@ const {execSync} = require('child_process'),
       '--resources', path.resolve(lib, 'resources'),
       args.alone ? '--alone' : '',
       args.tests ? '--tests' : ''
-    ];
+    ].filter((flag) => flag !== '');
   },
 
   /**
    * Wrapper for eo2js.
-   * @param {String} command - Command to execute
+   * @param {String|Array.<String>} command - eo2js sub-command, or a
+   *   sub-command plus its own positional arguments (e.g. `['dataize', obj]`)
    * @param {Object} args - Command arguments
    * @return {Promise<Array.<String>>}
    */
   eo2jsw = function(command, args) {
     const lib = path.resolve(__dirname, '../node_modules/eo2js/src'),
-      bin = path.resolve(lib, 'eo2js.js');
+      bin = path.resolve(lib, 'eo2js.js'),
+      cmd = Array.isArray(command) ? command : [command];
     return new Promise((resolve, reject) => {
-      execSync(
-        `node ${bin} ${command} ${flags(args, lib).filter((flag) => flag !== '').join(' ')}`,
-        {
-          timeout: 1200000,
-          windowsHide: true,
-          stdio: 'inherit'
-        }
-      );
-      resolve(args);
+      try {
+        execFileSync(
+          process.execPath,
+          [bin, ...cmd, ...flags(args, lib)],
+          {timeout: 1200000, windowsHide: true, stdio: 'inherit'}
+        );
+        resolve(args);
+      } catch (error) {
+        reject(error);
+      }
     });
   };
 
--- a/src/commands/js/dataize.js
+++ b/src/commands/js/dataize.js
@@ -14,7 +14,7 @@ const eo2jsw = require('../../eo2jsw');
  */
 module.exports = function(obj, args, opts) {
   return eo2jsw(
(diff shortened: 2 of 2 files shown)


ImilB:
@yegor256 please, take a look

yegor256:
@ImilB 
Thanks for your contribution! Unfortunately, we can't merge it without tests. Every feature we add and every bug we fix must be accompanied by at least one new test that confirms that the code was not working before. Without such a test, how can we know that what you commit actually works?

src/eo2jsw.js
```diff
@@ -20,28 +20,31 @@ const {execSync} = require('child_process'),
       '--resources', path.resolve(lib, 'resources'),
       args.alone ? '--alone' : '',
       args.tests ? '--tests' : ''
-    ];
+    ].filter((flag) => flag !== '');
```
Thayorns, line 23: `flags()` still returns `'--foreign eo-foreign.json'` as a single string on line 19. Under `execSync` the shell split that into two words before eo2js saw it; under `execFileSync` it arrives as one argv entry, so the parser is handed a flag literally named `--foreign eo-foreign.json` and the value never reaches it. That is the same defect this pull request is fixing, one line above the filter you added, and the description says it was split when it was not.

src/eo2jsw.js
```diff
@@ -20,28 +20,31 @@ const {execSync} = require('child_process'),
...
-   * @param {String} command - Command to execute
+   * @param {String|Array.<String>} command - eo2js sub-command, or a
+   *   sub-command plus its own positional arguments (e.g. `['dataize', obj]`)
    * @param {Object} args - Command arguments
    * @return {Promise<Array.<String>>}
    */
   eo2jsw = function(command, args) {
     const lib = path.resolve(__dirname, '../node_modules/eo2js/src'),
-      bin = path.resolve(lib, 'eo2js.js');
+      bin = path.resolve(lib, 'eo2js.js'),
```
Thayorns, line 35: `eo2jsw` now branches on the shape of its own argument, and its callers are split across both shapes: link.js:20, test.js:14 and transpile.js:18 hand it a string, dataize.js:16 hands it an array. Whoever writes the fifth caller has to guess which of the two is meant, and the wrapper keeps a branch that exists only because three call sites were left as they were.

Thayorns (commented):
Swaps `execSync` for `execFileSync` in the eo2js wrapper and passes the sub-command together with its positional arguments as argv entries, which takes the shell out of the path an object name travels. The direction is right and the place is right, but one flag in `flags()` is still a joined string and so falls into exactly the hole this pull request is filling. Nothing under test/ exercises `eo2jsw`, so neither the old quoting bug nor the new argv shape is pinned by a test.

0crat:
@imilb Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (33) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+12](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+4449](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1151, opened 2026-08-01
fix(#1146): correct the hasOwnProperty check in tinted-console.enable()

### LINKED ISSUE #1146: tinted-console.js: enable(level) erroneously enables unrelated levels and fails to flag unknown level names
##Summary

The if condition tests levels.hasOwnProperty(level) — the argument, which never changes across loop iterations — instead of levels.hasOwnProperty(key), the current key. Since level is a constant for the whole loop, the condition is either true for every iteration up to and including the target key, or false for all of them.
```js
module.exports.enable = function enable(level) {
  for (const key in levels) {
      levels[key] = true;
      if (key === level) {
        break;
      }
    }
  }
};
```
##Steps to reproduce
```js
const tinted = require('./src/tinted-console');
tinted.enable('debug');
```
(this is exactly what eoc.js does when --verbose/--debug is passed)

##Actual behavior

enable('debug') also flips trace to true, even though only debug was requested — because trace is iterated before debug in for...in order and the (constant) condition is already true.

Also, enable('some-typo') silently does nothing at all — no error, no warning — because the condition is false for every key, so the loop body never executes even once, not even for a name that doesn't exist in levels.

##Expected behavior
enable(level) should turn on exactly the requested level, and reject/flag an unknown level name instead of silently no-op'ing.

ImilB:
### Description

Removes the broken `for...in` loop entirely. The function now checks directly whether the requested level exists as a property of the `levels` object. If it doesn't, an error is thrown immediately with the unknown level name in the message. If it does, only that single key is set to `true`. The JSDoc is updated to document the new `@throws` behaviour and to clarify that only the requested level is enabled.

### What changed

**`src/tinted-console.js`** — The `enable` function body went from a multi-iteration loop with a mis-scoped condition to a straightforward guard-and-set: first check `!levels.hasOwnProperty(level)` and throw if the level is unknown, then `levels[level] = true` directly. This eliminates the loop, the incorrect variable reference, and the silent failure on typos all at once.
Fixes #1146

### DIFF
--- a/src/tinted-console.js
+++ b/src/tinted-console.js
@@ -39,16 +39,13 @@ for (const level in levels) {
 }
 
 /**
- * Enable this particular logging level.
+ * Enable this particular logging level (and only this one).
  * @param {String} level - The level to enable
+ * @throws {Error} If the level name is not one of the known levels
  */
 module.exports.enable = function enable(level) {
-  for (const key in levels) {
-    if (levels.hasOwnProperty(level)) {
-      levels[key] = true;
-      if (key === level) {
-        break;
-      }
-    }
+   if (!levels.hasOwnProperty(level)) {
+    throw new Error(`Unknown log level: "${level}"`);
   }
+  levels[level] = true;
 };


Thayorns (commented):
Replaces the `for...in` loop in `enable` with a direct membership check and a single assignment, which is the right shape: the mis-scoped condition is gone, an unknown level now fails loudly instead of doing nothing, and `enable('debug')` no longer drags `trace` on with it. One indentation slip will stop `npx eslint`. There is no test for either of the two behaviours the fix introduces, so the file only keeps its existing `console.warn` smoke test and nothing pins the throw or the single-level effect.

src/tinted-console.js
```diff
@@ -39,16 +39,13 @@ for (const level in levels) {
...
  */
 module.exports.enable = function enable(level) {
-  for (const key in levels) {
-    if (levels.hasOwnProperty(level)) {
-      levels[key] = true;
-      if (key === level) {
-        break;
-      }
-    }
+   if (!levels.hasOwnProperty(level)) {
```
Thayorns, line 47: This line is indented three spaces while the closing brace under it sits at two. `eslint.config.js` sets `'indent': ['error', 2, {"SwitchCase": 1}]`, so `npx eslint` reports this file as an error and the build stops before anything else in the change is looked at.

yegor256:
CI is failing on this PR: the `eslint` check is red. Please fix the linting errors before this can be merged.

---

yegor256:
CI is failing: the `eslint` check is red (from Aug 1). Could you please fix this failure before this can be merged?

---

yegor256:
eslint plus build/itest are red on this PR. Not merging until CI is green.

---

PR #1152, opened 2026-08-01
fix(#1147): use a relative, forward-slash CSS href instead of an absolute path

### LINKED ISSUE #1147: Generated docs (eoc docs) with absolute css path fail to load stylesheet on Windows and expose local paths
### Summary
n src/commands/docs.js:

```s
...
fs.writeFileSync(html_app, wrapHtml(name, xmir_html, css));
...
fs.writeFileSync(info.path, generatePackageHtml(`${package_name} package`, info.xmir_htmls, css));
```
css is computed with path.join(output, ...) where output = path.resolve(opts.target, 'docs') — an absolute filesystem path — and it's the same value for every generated HTML file regardless of that file's own location. This path is inserted directly as the href:
```s
```
### Actual behavior

On Windows, path.join produces backslashes, e.g. C:\Users\alice\project\.eoc\docs\styles.css. That's not a valid href/URL — browsers don't resolve backslash paths — so every generated page loses its stylesheet. Even on Unix, the generated HTML embeds the author's absolute local directory structure into files meant to be shared/published (e.g. via GitHub Pages), which is unnecessary information disclosure and breaks the moment the docs/ folder is moved or copied elsewhere.

ImilB:
### What changed

Added a `cssHref(fromDir, cssPath)` helper that computes the relative path from any HTML file's directory to `styles.css` and normalises separators to forward slashes. All three call sites that write HTML now use this helper, passing the directory of the file being written so each gets a relative path matching its depth. `wrapHtml` and `generatePackageHtml` signatures are unchanged — they still receive a string, now correctly relative.

Fixes #1147

### DIFF
--- a/src/commands/docs.js
+++ b/src/commands/docs.js
@@ -1,4 +1,4 @@
-/*
+  /*
  * SPDX-FileCopyrightText: Copyright (c) 2022-2026 Objectionary.com
  * SPDX-License-Identifier: MIT
  */
@@ -121,6 +121,16 @@ function wrapHtml(name, html, css) {
  * @param {Hash} opts - All options
  * @return {Promise<String>} Resolves to the message reporting the summary path
  */
+/**
+ * Build a browser-safe relative href from an HTML file's directory to the
+ * shared stylesheet, always using forward slashes regardless of OS.
+ * @param {String} fromDir - Directory of the HTML file that will use it
+ * @param {String} cssPath - Absolute path of styles.css
+ * @return {String} Relative href
+ */
++function cssHref(fromDir, cssPath) {
++  return path.relative(fromDir, cssPath).split(path.sep).join('/');
++}
 module.exports = function(opts) {
   return elapsed(async (tracked) => {
     try {
@@ -138,7 +148,10 @@ module.exports = function(opts) {
         const xmir_html = createXmirHtmlBlock(xmir);
         const html_app = path.join(output, path.dirname(relative),`${name}.html`);
         fs.mkdirSync(path.dirname(html_app), {recursive: true});
-        fs.writeFileSync(html_app, wrapHtml(name, xmir_html, css));
+        fs.writeFileSync(
+          html_app,
+          wrapHtml(name, xmir_html, cssHref(path.dirname(html_app), css))
+        );
         const packages = path.dirname(relative).split(path.sep).join('.');
         const html_package = path.join(output, `package_${packages}.html`);
         if (!packages_info.has(packages)) {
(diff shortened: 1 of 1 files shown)


src/commands/docs.js
```diff
@@ -121,6 +121,16 @@ function wrapHtml(name, html, css) {
...
  * @return {Promise<String>} Resolves to the message reporting the summary path
  */
+/**
+ * Build a browser-safe relative href from an HTML file's directory to the
+ * shared stylesheet, always using forward slashes regardless of OS.
+ * @param {String} fromDir - Directory of the HTML file that will use it
+ * @param {String} cssPath - Absolute path of styles.css
+ * @return {String} Relative href
+ */
++function cssHref(fromDir, cssPath) {
```
Thayorns, line 131: Lines 131 to 133 begin with a literal `+` in the committed file, not just in the diff view. `+function cssHref(...) {` parses as unary plus on a function expression, and the very next line, `+  return ...`, is a syntax error because `return` cannot follow a `+`. `require('./commands/docs')` throws at load, so `eoc docs` cannot start at all. It looks like a patch was pasted into the file rather than applied to it.

src/commands/docs.js
```diff
@@ -121,6 +121,16 @@ function wrapHtml(name, html, css) {
  * @param {Hash} opts - All options
  * @return {Promise<String>} Resolves to the message reporting the summary path
  */
+/**
```
Thayorns, line 124: The "Command to generate documentation" block on lines 119 to 123 used to sit on `module.exports`. The new block and `cssHref` are now wedged between the two, so the command is left undocumented and `cssHref` carries two docblocks, the first of which describes something else entirely.

src/commands/docs.js
```diff
@@ -1,4 +1,4 @@
-/*
+  /*
```
Thayorns, line 1: Two spaces crept in before the SPDX header, so the comment no longer starts at column one. Nothing in the change needed to touch this line.

Thayorns (commented):
Introduces `cssHref` so every HTML file gets a stylesheet link relative to its own directory instead of an absolute filesystem path, and routes all three writers through it. The idea is right and the three call sites are all covered. However, the committed file carries raw diff markers in it, so `src/commands/docs.js` no longer parses, and no test in test/commands/test_docs.js asserts anything about the href that is actually written into the HTML.

yegor256:
CI is failing on this PR: `itest (macos-15, 19, 18, Java)`, `build (ubuntu-24.04, 22, 18)`, `nix`, and `eslint` checks are all red. Please fix the failures before this can be merged.

---

yegor256:
CI is failing: `itest (macos-15, 19, 18, Java)`, `build (ubuntu-24.04, 22, 18)`, `nix`, and `eslint` checks are red (from Aug 1). Could you please fix these failures before this can be merged?

---

yegor256:

---

yegor256:
Please forgive us for closing this pull request: it has merge conflicts with the current master branch and has been open for more than a month, so it does not meet our bar for staying open.

This is not a judgment on the contribution itself, just routine stale-PR cleanup. We would be glad if you resubmit a new pull request, perhaps similar to this one, but rebased on top of the latest master branch. Thank you for your effort and understanding.

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (3) comments; -4 for too few (26) hits-of-code; +1 to give you at least something.  Your running score is [+4454](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1155, opened 2026-08-02
#1121: declare saxon-js as direct dependency

### LINKED ISSUE #1121: `saxon-js` dependency not declared in `package.json` causing `Cannot find module` error
## What happens

`src/commands/docs.js:8` requires `saxon-js` directly:

```js
const SaxonJS = require('saxon-js');
```

`saxon-js` is not listed in `package.json`'s `dependencies`. It currently resolves only because it happens to be pulled in transitively through `eo2js` (`eo2js@0.0.8 → saxon-js@2.7.0`), confirmed with `npm ls saxon-js`.

By contrast, `marked`, which is also used only by `docs.js`, is correctly declared as a direct dependency — so the two modules `docs.js` depends on are handled inconsistently.

Verified: removing `node_modules/saxon-js` and running `eoc docs` produces a raw, uncaught error:
```
Error: Cannot find module 'saxon-js'
```
with a full stack trace and exit code 1 — not a controlled CLI error message.

Because the dependency is undeclared, it can silently disappear in three realistic scenarios: a future `eo2js` version drops or renames its own `saxon-js` dependency, a package manager that enforces strict `node_modules` isolation (e.g. `pnpm`, or npm with `--legacy-peer-deps` off in some configurations) refuses to hoist a transitive dependency this way, or `eo2js` itself is ever swapped out.

## What should happen

`saxon-js` should be added to `package.json`'s own `dependencies`, matching how `marked` is already declared, so `eoc docs` does not depend on an implementation detail of a different dependency's own dependency tree.

## Steps to reproduce

1. In a checkout of `eoc`, run `npm install` then `rm -rf node_modules/saxon-js`.
2. Run `node src/eoc.js docs`.
3. Observe an uncaught `Error: Cannot find module 'saxon-js'` instead of `docs` working (since `eo2js`, the package that actually depends on it, is still installed and unaffected).

morphqdd:
Closes #1121

`saxon-js` is required directly by `src/commands/docs.js` but was not declared in `package.json`'s own `dependencies`, only pulled in transitively through `eo2js`. Added it as a direct dependency, matching how `marked` (also used only by `docs.js`) is already declared.

Added a test asserting `saxon-js` is listed in `package.json` dependencies.

### DIFF
--- a/test/commands/test_docs.js
+++ b/test/commands/test_docs.js
@@ -370,4 +370,13 @@ describe('docs', () => {
     );
     done();
   });
+  /**
+   * Tests that 'saxon-js', required directly by this module, is declared
+   * as its own dependency instead of relying on it being pulled in
+   * transitively through 'eo2js'.
+   */
+  it('declares saxon-js as a direct dependency', () => {
+    const pkg = JSON.parse(fs.readFileSync(path.join(__dirname, '..', '..', 'package.json'), 'utf-8'));
+    assert.ok(pkg.dependencies['saxon-js'], 'saxon-js must be listed in package.json dependencies');
+  });
 });
--- a/test/commands/test_docs.js
+++ b/test/commands/test_docs.js
@@ -371,12 +371,22 @@ describe('docs', () => {
     done();
   });
   /**
-   * Tests that 'saxon-js', required directly by this module, is declared
-   * as its own dependency instead of relying on it being pulled in
-   * transitively through 'eo2js'.
+   * Tests that every package 'docs.js' requires is declared by 'eoc' itself
+   * and resolves, instead of being reached transitively through 'eo2js'.
    */
-  it('declares saxon-js as a direct dependency', () => {
+  it('declares every package it requires', () => {
+    const source = path.join(__dirname, '..', '..', 'src', 'commands', 'docs.js');
     const pkg = JSON.parse(fs.readFileSync(path.join(__dirname, '..', '..', 'package.json'), 'utf-8'));
-    assert.ok(pkg.dependencies['saxon-js'], 'saxon-js must be listed in package.json dependencies');
+    const required = [...fs.readFileSync(source, 'utf-8').matchAll(/require\('(?<name>[^.'][^']*)'\)/g)]
+      .map((match) => match.groups.name)
+      .filter((name) => !require('module').builtinModules.includes(name));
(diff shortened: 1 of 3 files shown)


morphqdd:

test/commands/test_docs.js
```diff
@@ -370,4 +370,13 @@ describe('docs', () => {
     );
     done();
   });
+  /**
+   * Tests that 'saxon-js', required directly by this module, is declared
+   * as its own dependency instead of relying on it being pulled in
+   * transitively through 'eo2js'.
+   */
+  it('declares saxon-js as a direct dependency', () => {
+    const pkg = JSON.parse(fs.readFileSync(path.join(__dirname, '..', '..', 'package.json'), 'utf-8'));
```
Thayorns, line 379: This reads `package.json` and asserts that the key the same commit added is present, so it passes for as long as nobody edits that one line, and it never loads `src/commands/docs.js` or `saxon-js` at all. The breakage in #1121 was a module that would not resolve, and this test stays green through it: strip `saxon-js` out of `eo2js` upstream, leave the manifest untouched, and `docs` fails on `require('saxon-js')` while the suite reports nothing wrong.

Thayorns (commented):
Declares `saxon-js` in `dependencies` so `src/commands/docs.js` stops relying on it being hoisted out of `eo2js`, which is the right call and matches how `marked` is already handled. The accompanying test guards the manifest line rather than the resolution it is there to protect, so the failure mode the issue describes can come back without turning the suite red.

morphqdd:

morphqdd:
@yegor256, please review

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (21) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6703](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +4 points for this: +12 as a basis; -5 for very few (1) comments; -4 for too few (21) hits-of-code; +1 to give you at least something.  Your running score is [+4453](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1156, opened 2026-08-02
#500: start the inspection server and ask it where the session begins

morphqdd:
First slice of #500. It starts the server, asks it one question and prints the answer. The verbs come next, each marked with a `@todo` in the code.

- `inspect/` is a small Maven project that builds `inspect.jar` with `org.eolang.eoc.Inspect` inside. It runs a Takes server on the port given by `--port` and answers `GET /` with JSON about the object the session starts at. The root object is taken from memory, nothing is read from the classpath or from XMIR. `eo-runtime` is `provided`, since the program's own jar already carries it.
- `src/commands/java/inspect.js` builds that jar on first use, spawns `java -cp <target>/eoc.jar:<target>/inspect/inspect.jar org.eolang.eoc.Inspect --port N`, waits for the port, asks, prints, and kills the JVM.
- `eoc inspect` is registered in `src/eoc.js`, with `--port` defaulting to 8080.

This follows the design in your comment above: a second jar, the EO jar untouched, a background JVM, Takes, JSON back to the JavaScript side.

Two tests. One compiles a program and checks the command prints the root object. The other runs the JavaScript side against a stub server and checks both jars land on the classpath and the JVM is killed afterwards.

Not closing the issue with this PR. The puzzles say what is left: loading the whole tree at startup, then the verbs in four groups, then the interactive loop and killing the JVM on signals.

### DIFF
--- a/inspect/pom.xml
+++ b/inspect/pom.xml
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2022-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.eolang</groupId>
+  <artifactId>eoc-inspect</artifactId>
+  <version>0.0.0</version>
+  <packaging>jar</packaging>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <eo.version>undefined</eo.version>
+    <eo.targetDir>${project.basedir}/target</eo.targetDir>
+    <takes.version>1.24.6</takes.version>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>org.takes</groupId>
+      <artifactId>takes</artifactId>
+      <version>${takes.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.eolang</groupId>
+      <artifactId>eo-runtime</artifactId>
+      <version>${eo.version}</version>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+  <build>
+    <finalName>inspect</finalName>
+    <directory>${eo.targetDir}/inspect</directory>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.8.1</version>
+        <configuration>
+          <release>11</release>
+        </configuration>
+      </plugin>
+      <plugin>
+        <artifactId>maven-shade-plugin</artifactId>
+        <version>3.6.2</version>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <createDependencyReducedPom>false</createDependencyReducedPom>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                  <mainClass>org.eolang.eoc.Inspect</mainClass>
+                </transformer>
+              </transformers>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+</project>
--- a/inspect/src/main/java/org/eolang/eoc/Inspect.java
+++ b/inspect/src/main/java/org/eolang/eoc/Inspect.java
@@ -1,0 +1,118 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+
+package org.eolang.eoc;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import org.eolang.Phi;
+import org.takes.Take;
+import org.takes.facets.fork.FkRegex;
+import org.takes.facets.fork.TkFork;
+import org.takes.http.Exit;
+import org.takes.http.FtBasic;
+import org.takes.rs.RsText;
+import org.takes.rs.RsWithType;
+
+/**
+ * The server behind the {@code eoc inspect} command.
+ *
+ * <p>It is started by the JavaScript side as a background process, with
+ * the linked program on the classpath, and it answers HTTP requests about
+ * the tree of objects that lives in its own memory. Nothing is read from
+ * the classpath or from XMIR: the objects are asked directly.</p>
+ *
+ * @todo #500:60min Load the whole tree of objects at startup.
+ *  Right now the only object this server can reach is the root, and it
+ *  reaches it lazily through {@link Phi}. A package object learns that a
+ *  child exists only when that child is taken by name, so listing
+ *  everything under the root is impossible until every object is loaded
+ *  into memory once, at startup, the way the session in this issue shows
+ *  with its "Loaded 56 objects" line. Load them here and keep them, so
+ *  that the verbs below walk a tree that is already there.
+ * @since 0.0.0
+ */
+public final class Inspect {
+
+    /**
+     * Command line arguments the server was started with.
+     */
+    private final List<String> flags;
+
+    /**
+     * Ctor.
+     * @param args Command line arguments
+     */
+    public Inspect(final String... args) {
+        this.flags = Arrays.asList(args);
+    }
+
+    /**
+     * Entry point.
+     * @param args Command line arguments
+     * @throws IOException If fails
+     */
+    public static void main(final String... args) throws IOException {
+        new Inspect(args).start();
+    }
+
+    /**
+     * Start the server and serve until the process is killed.
+     *
+     * @todo #500:90min Answer the verbs that only read the tree.
+     *  The session in this issue moves around before it changes anything:
+     *  ls prints the attributes of the object we are at, go (or .foo)
+     *  steps into one of them, and .. steps back out. All three need the
+     *  server to remember where the session currently is, so add that
+     *  position here first and let these three verbs move it.
+     * @todo #500:90min Answer the verbs that change the current object.
+     *  Here add (or +foo) puts a new void attribute on the object we are
+     *  at, rm (or -foo) takes one away, form attaches an empty formation,
+     *  put attaches data to the delta asset, and to attaches an object
+     *  named elsewhere. All five write into the object, which is what the
+     *  base class of objectionary/eo#6273 is meant to allow.
+     * @todo #500:60min Answer the verbs that copy and dispatch.
+     *  Here cp makes a copy of the object we are at and attaches it under
+     *  a name the caller chooses, and dd dispatches one attribute and
+     *  attaches the result the same way. Both need somewhere to keep
+     *  those names, since a name like the one in the session is invented
+     *  by the user and belongs to no object of the program.
+     * @todo #500:30min Answer the dataize verb.
+     *  Here dataize (or run) dataizes the object the session is at and
+     *  answers with its bytes, which the JavaScript side prints the way
+     *  the session in the issue shows. A program that fails to dataize is
+     *  the reason this tool exists, so answer with the failure instead of
+     *  letting the server die on it.
+     * @throws IOException If fails
+     */
+    public void start() throws IOException {
+        new FtBasic(
+            new TkFork(
+                new FkRegex(
+                    "/",
+                    (Take) req -> new RsWithType.Json(
+                        new RsText(String.format("{\"forma\":\"%s\"}", Phi.Φ.forma()))
+                    )
+                )
+            ),
+            this.port()
+        ).start(Exit.NEVER);
+    }
+
+    /**
+     * The port to listen on, taken from the {@code --port} option.
+     * @return The port number
+     */
+    private int port() {
+        final int idx = this.flags.indexOf("--port");
+        if (idx < 0 || idx + 1 == this.flags.size()) {
+            throw new IllegalArgumentException(
+                "The --port option is mandatory, as in \"--port 8080\""
+            );
+        }
+        return Integer.parseInt(this.flags.get(idx + 1));
+    }
+}
--- a/src/commands/java/inspect.js
+++ b/src/commands/java/inspect.js
@@ -1,0 +1,97 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+
+const fs = require('fs');
+const path = require('path');
+const {spawn} = require('node:child_process');
+const {mvnw, flags} = require('../../mvnw');
+const {verifyJavac} = require('../../jdk');
+
+/**
+ * The jar with the inspection server, built on demand.
+ * @param {Object} opts - All options
+ * @return {Promise<String>} Path to the jar
+ */
+async function jar(opts) {
+  const built = path.resolve(opts.target, 'inspect', 'inspect.jar');
+  if (!fs.existsSync(built)) {
+    await mvnw(
+      ['package', '-f', path.resolve(__dirname, '../../../inspect/pom.xml')].concat(flags(opts)),
+      opts.target,
+      opts.batch
+    );
+  }
+  return built;
+}
+
+/**
+ * Ask the server where the session currently is, waiting for it to open the port.
+ * The address is IPv4 on purpose, since the server listens on IPv4 while
+ * "localhost" resolves to ::1 first on some machines and is refused there.
+ * @param {Number} port - TCP port the server listens on
+ * @param {Number} deadline - Moment in time after which we give up, in millis
+ * @return {Promise<Object>} The answer, parsed from JSON
+ */
+async function ask(port, deadline) {
+  let answer;
+  try {
+    answer = await (await fetch(`http://127.0.0.1:${port}/`)).json();
+  } catch (error) {
+    if (Date.now() > deadline) {
+      throw new Error(
+        `The inspection server has not opened port ${port} in a minute`,
+        {cause: error}
+      );
+    }
+    await new Promise((resolve) => setTimeout(resolve, 200));
+    answer = await ask(port, deadline);
+  }
+  return answer;
+}
+
+/**
+ * Traverse the tree of objects of a compiled program.
+ *
+ * The server runs in a separate JVM, with the linked program on its
+ * classpath, and every question about the tree is an HTTP request to it.
+ *
+ * @todo #500:60min Turn this single question into an interactive session.
+ *  The issue asks for a prompt that stays open, reads a verb from the
+ *  terminal, sends it to the server and prints the answer, until the user
+ *  leaves. Right now we ask one question and shut the server down. Add
+ *  the loop, and keep the printing here, since the server is the only one
+ *  that knows the objects and this side is the only one that knows the
+ *  terminal.
+ * @todo #500:30min Do not leave the JVM behind when this process dies.
+ *  The server is killed in the "finally" below, which covers a failure of
+ *  our own code, but not a user pressing Ctrl-C and not a crash of the
+ *  Node process itself. In both cases the JVM keeps the port open and the
+ *  next run of the command fails to bind it. Kill the child on the
+ *  signals too, and report a port that is already taken by naming it.
+ * @param {Object} opts - All options
+ * @param {Function} [exec] - Optional command runner for the JDK check
+ * @param {Function} [runner] - Optional Java process runner
+ * @return {Promise} of the inspection session
+ */
+module.exports = async function(opts, exec, runner = spawn) {
+  verifyJavac(exec);
+  const port = Number(opts.port);
+  const params = [
+    '-cp',
+    [path.resolve(opts.target, 'eoc.jar'), await jar(opts)].join(path.delimiter),
+    'org.eolang.eoc.Inspect',
+    '--port',
+    String(port),
+  ];
+  console.debug(`+ java ${params.join(' ')}`);
+  const server = runner('java', params, {stdio: 'inherit'});
+  try {
+    const answer = await ask(port, Date.now() + 60000);
+    console.info('Ready to traverse the Universe');
+    console.info(`@ ${answer.forma}`);
+  } finally {
+    server.kill();
+  }
+};
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -48,6 +48,7 @@ const commands = {
       compile: require('./commands/java/compile'),
       dataize: require('./commands/java/dataize'),
       test: require('./commands/java/test'),
+      inspect: require('./commands/java/inspect'),
     }
   },
   [language.js]: {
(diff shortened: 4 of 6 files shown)


morphqdd:

test/commands/test_inspect.js
```diff
@@ -0,0 +1,100 @@
...
+    const home = path.resolve('temp/test-inspect');
+    fs.rmSync(home, {recursive: true, force: true});
+    fs.mkdirSync(home, {recursive: true});
+    fs.writeFileSync(
+      path.resolve(home, 'simple.eo'),
+      ['[] > simple', '  42 > @'].join('\n')
+    );
+    const stdout = runSync([
+      'inspect',
+      '--port=18081',
```
Thayorns: Port 18081 is fixed, and with `this.timeout(0)` on line 16 nothing bounds the wait. When the port is already taken, by a JVM the previous run left behind or by a second suite on the same machine, the server never binds, `ask` in inspect.js retries for its full minute, and the run hangs before it fails. The sibling test on line 49 already does the right thing with `server.listen(0, ...)`.

test/commands/test_inspect.js
```diff
@@ -0,0 +1,100 @@
...
+          return {kill: () => {
+            killed = true;
+          }};
+        }
+      );
+    } finally {
+      console.info = info;
+      server.close();
+    }
+    assert(
```
Thayorns, line 91: Five assertions here, on lines 70, 74, 78, 82 and 86, covering the class name, two classpath entries, the printed line and the kill. The first one that fails hides the other four, so a red run tells you one thing when it could tell you five, and the test has no single behaviour you could name it after.

test/commands/test_inspect.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+  it('fails fast with a clear message when javac is not on the PATH', async () => {
+    const missing = () => {
+      const cause = new Error('spawnSync javac ENOENT');
+      cause.code = 'ENOENT';
+      throw cause;
+    };
+    await assert.rejects(
+      () => inspect({target: '.', port: 8080}, missing),
+      /javac/,
```
Thayorns: The `/javac/` matcher ties this test to the wording of `missingJavacMessage` in src/jdk.js, which spells the word out on four of its six lines. Rephrasing any of that diagnostic turns this red without anything being broken, and what you want to know is that `verifyJavac` refused, not what it said while refusing.

test/commands/test_inspect.js
```diff
@@ -0,0 +1,100 @@
...
+      'inspect',
+      '--port=18081',
+      '--easy',
+      '--blind',
+      `--parser=${parserVersion}`,
+      `--home-tag=${homeTag}`,
+      '-s', home,
+      '-t', path.resolve(home, 'target'),
+    ]);
+    assert(stdout.includes('@ Φ'), stdout);
```
Thayorns: The failure message is the whole of stdout, so when this goes red the report is a wall of build output with no statement of what went wrong in it. The other four assertions in this file each name the failure, and this one reads oddly next to them.

Thayorns (commented):
First slice of the inspection session: a separate Maven module producing `inspect.jar`, a Takes server answering one question about the root object, a JavaScript side that builds the jar on demand, spawns the JVM, asks and kills it, and an `eoc inspect` command wired into the pipeline. The split is sound, the puzzles say honestly what is missing, and the JavaScript side is tested against a stub rather than a mock. The findings below are all in the tests: the integration test binds a fixed port, and the unit test carries five assertions for what reads as one behaviour.

inspect/src/main/java/org/eolang/eoc/Inspect.java
```diff
@@ -0,0 +1,118 @@
...
+     *  letting the server die on it.
+     * @throws IOException If fails
+     */
+    public void start() throws IOException {
+        new FtBasic(
+            new TkFork(
+                new FkRegex(
+                    "/",
+                    (Take) req -> new RsWithType.Json(
+                        new RsText(String.format("{\"forma\":\"%s\"}", Phi.Φ.forma()))
```
Favixx, line 97: Hand-built JSON. `Φ.forma()` is safe today, so nothing is broken right now — but the four `@todo`s below are all about answering with objects the user names (`cp` "attaches it under a name the caller chooses", `put` "attaches data to the delta asset"), and the moment one of those names or one of those byte strings carries a `"` or a `\`, this response stops being JSON and `ask()` on the other side fails in the confusing way described on line 41.

Takes already carries a JSON response — `RsJson` over a `javax.json` object — which escapes and also removes the need for `RsWithType.Json` to declare a type the body may not have. It is one import and three lines now; after the verbs land, it is a rewrite of five responses.

inspect/src/main/java/org/eolang/eoc/Inspect.java
```diff
@@ -0,0 +1,118 @@
...
+     * @return The port number
+     */
+    private int port() {
+        final int idx = this.flags.indexOf("--port");
+        if (idx < 0 || idx + 1 == this.flags.size()) {
+            throw new IllegalArgumentException(
+                "The --port option is mandatory, as in \"--port 8080\""
+            );
+        }
+        return Integer.parseInt(this.flags.get(idx + 1));
```
Favixx, line 116: The two lines above go to real trouble to explain a missing `--port` ("The --port option is mandatory, as in \"--port 8080\""), and then this one lets `--port eight` out as a bare `NumberFormatException: For input string: "eight"` with no mention of `--port` at all. Worth extending the same check: catch the parse failure and re-throw with the option named.

There is a second value worth rejecting here. A port outside 1..65535 fails deep inside `FtBasic` with a message about a socket; caught here it can say which option was wrong.

inspect/src/main/java/org/eolang/eoc/Inspect.java
```diff
@@ -0,0 +1,118 @@
...
+            new TkFork(
+                new FkRegex(
+                    "/",
+                    (Take) req -> new RsWithType.Json(
+                        new RsText(String.format("{\"forma\":\"%s\"}", Phi.Φ.forma()))
+                    )
+                )
+            ),
+            this.port()
+        ).start(Exit.NEVER);
```
Favixx, line 102: `Exit.NEVER` makes killing the process the only way to stop this server, which is what the `@todo` on the JavaScript side is about — but that puzzle is scoped to catching signals, and it will not be enough on its own. `server.kill()` sends a signal and returns; the JVM then unwinds, and the listening socket is released some time after the Node process has already exited. Two `eoc inspect` runs in quick succession on the default 8080 can therefore collide even when nothing crashed and nothing was interrupted.

An `/exit` fork answered with a shutdown, asked for before `kill()`, gives the server a chance to close the socket itself. If that is out of scope for this slice, it is worth adding to the 30min puzzle, because "report a port that is already taken by naming it" treats the symptom while this is one of the causes.

src/commands/java/inspect.js
```diff
@@ -0,0 +1,97 @@
...
+const {verifyJavac} = require('../../jdk');
+
+/**
+ * The jar with the inspection server, built on demand.
+ * @param {Object} opts - All options
+ * @return {Promise<String>} Path to the jar
+ */
+async function jar(opts) {
+  const built = path.resolve(opts.target, 'inspect', 'inspect.jar');
+  if (!fs.existsSync(built)) {
```
Favixx, line 19: Existence is being used as a proxy for freshness, and it is a poor one. After the first run, `inspect.jar` is on disk forever: editing `Inspect.java`, or checking out a branch where the server answers differently, changes nothing until somebody deletes the file by hand. Since the four `@todo`s in `Inspect.java` describe most of the eventual behaviour of this command, the people hit by this are the ones implementing them.

The cheap fix is to compare the mtime of the jar against the newest file under `inspect/src` and rebuild when the sources win. The cheaper one is to let Maven decide — `package` on an up-to-date module is fast — and keep the `existsSync` only as an offline fallback.

src/commands/java/inspect.js
```diff
@@ -0,0 +1,97 @@
...
+    await mvnw(
+      ['package', '-f', path.resolve(__dirname, '../../../inspect/pom.xml')].concat(flags(opts)),
+      opts.target,
+      opts.batch
+    );
+  }
+  return built;
```
Favixx, line 26: `built` is returned whether or not the build put anything there. If `mvnw` succeeds but the artifact lands elsewhere — a changed `finalName`, a changed `<directory>`, a shade execution that did not run — this returns a path to a file that does not exist, and the failure surfaces two steps later as a `NoClassDefFoundError` from the JVM, pointing at the classpath rather than at the build.

The pom has both `<finalName>inspect</finalName>` and `<directory>${eo.targetDir}/inspect</directory>`, and this path re-derives the same convention independently in JavaScript. An `fs.existsSync(built)` check after the build, throwing with the path in the message, keeps the two in step.

src/commands/java/inspect.js
```diff
@@ -0,0 +1,97 @@
...
+ * "localhost" resolves to ::1 first on some machines and is refused there.
+ * @param {Number} port - TCP port the server listens on
+ * @param {Number} deadline - Moment in time after which we give up, in millis
+ * @return {Promise<Object>} The answer, parsed from JSON
+ */
+async function ask(port, deadline) {
+  let answer;
+  try {
+    answer = await (await fetch(`http://127.0.0.1:${port}/`)).json();
+  } catch (error) {
```
Favixx, line 41: This `catch` covers `fetch` and `.json()` together, so it cannot tell "nothing is listening yet" from "something answered and the answer is not JSON". The first is the condition this retry loop exists for; the second is a bug in the server that will be retried two hundred and ninety-nine times and then reported as `The inspection server has not opened port N in a minute`.

Given the `String.format` on `Inspect.java:97`, this is not hypothetical. Awaiting the response first and parsing outside the retry — retry on a connection error, fail immediately on a bad body — makes each failure say what it is. A non-2xx status deserves the same treatment: right now a 500 from a broken take is also a "port not opened".

src/commands/java/inspect.js
```diff
@@ -0,0 +1,97 @@
...
+    if (Date.now() > deadline) {
+      throw new Error(
+        `The inspection server has not opened port ${port} in a minute`,
+        {cause: error}
+      );
+    }
+    await new Promise((resolve) => setTimeout(resolve, 200));
+    answer = await ask(port, deadline);
```
Favixx, line 49: Recursion for a retry loop that runs for a minute at 200ms a turn: three hundred frames deep by the time it gives up, all of them live, each holding an `error` from its own attempt. It works, and V8 will not overflow at that depth, but a `while (Date.now() < deadline)` says the same thing in the same number of lines and keeps the stack flat.

It would also fix what the deadline currently reports. The `cause` attached on line 45 is the error of whichever attempt happened to be the last one — a loop can keep the first, which is usually the one that says why the server never came up.

src/commands/java/inspect.js
```diff
@@ -0,0 +1,97 @@
...
+ *  next run of the command fails to bind it. Kill the child on the
+ *  signals too, and report a port that is already taken by naming it.
+ * @param {Object} opts - All options
+ * @param {Function} [exec] - Optional command runner for the JDK check
+ * @param {Function} [runner] - Optional Java process runner
+ * @return {Promise} of the inspection session
+ */
+module.exports = async function(opts, exec, runner = spawn) {
+  verifyJavac(exec);
+  const port = Number(opts.port);
```
Favixx, line 80: `Number(undefined)` and `Number('eight')` are both `NaN`, and `String(NaN)` is `'NaN'`, which is what gets passed to `--port` and blows up inside `Integer.parseInt` in the server — a stack trace from another process for a typo on the command line.

Commander gives `--port` a default of `'8080'`, so `undefined` only happens when this module is called directly, which is exactly what `test_inspect.js` does. Rejecting a non-integer or an out-of-range port here, before the JVM is spawned, costs three lines and means neither side has to guess what the other checked.

src/commands/java/inspect.js
```diff
@@ -0,0 +1,97 @@
...
+  const params = [
+    '-cp',
+    [path.resolve(opts.target, 'eoc.jar'), await jar(opts)].join(path.delimiter),
+    'org.eolang.eoc.Inspect',
+    '--port',
+    String(port),
+  ];
+  console.debug(`+ java ${params.join(' ')}`);
+  const server = runner('java', params, {stdio: 'inherit'});
```
Favixx, line 89: `stdio: 'inherit'` sends everything the JVM writes straight to the user's terminal, unbuffered and unlabelled: Takes' own startup logging, any `slf4j` output from `eo-runtime`, and every stack trace the four future verbs produce, all interleaved with the `console.info` lines this command prints.

For one question and one answer it is merely untidy. For the interactive prompt the first `@todo` describes — a loop reading a verb and printing an answer — output arriving from a second process between the prompt and the reply will make the session hard to read. Capturing the child's streams and either discarding them or printing them under `--verbose` is worth doing while there is only one caller.

src/commands/java/inspect.js
```diff
@@ -0,0 +1,97 @@
...
+  try {
+    const answer = await ask(port, Date.now() + 60000);
+    console.info('Ready to traverse the Universe');
+    console.info(`@ ${answer.forma}`);
+  } finally {
+    server.kill();
```
Favixx, line 95: `kill()` requests termination and returns immediately, so this function resolves while the JVM is still shutting down and still holding the port. The next `eoc inspect` — same default 8080 — can then fail to bind for a reason that has nothing to do with the user.

Awaiting the child's `exit` event, with a short timeout followed by `kill('SIGKILL')`, makes the command's completion mean the port is free. That is also what the `@todo` above needs in order to keep its promise about naming a port that is already taken, since otherwise the most common taker is the previous run of this same command.

inspect/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.eolang</groupId>
+  <artifactId>eoc-inspect</artifactId>
+  <version>0.0.0</version>
+  <packaging>jar</packaging>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <eo.version>undefined</eo.version>
```
Favixx, line 15: `undefined` as the default of `eo.version` is fine when `flags()` always overrides it — which it does, with `-Deo.version=${opts.parser}`. It is not fine for the person who does the obvious thing and runs `mvn -f inspect/pom.xml package` by hand while working on the four puzzles in `Inspect.java`: Maven then tries to resolve `org.eolang:eo-runtime:undefined` and says so in fifty lines of dependency resolution.

A default of `LATEST` would be worse. Either the `maven-enforcer-plugin` with a `requireProperty` rule that names the flag to pass, or a comment on this line pointing at `src/mvnw.js`, saves that person ten minutes.

inspect/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+  <build>
+    <finalName>inspect</finalName>
+    <directory>${eo.targetDir}/inspect</directory>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.8.1</version>
```
Favixx, line 38: Two plugin versions hardcoded here while `takes.version` gets a property one screen up — worth being consistent, especially since `maven-compiler-plugin` 3.8.1 is from 2019 and is the one version in this file that a dependency bot will not find behind a property.

`<release>11</release>` deserves a second look too. The runtime this jar has to load next to is `eo-runtime`, and it runs in whatever JVM the user has; compiling the server at 11 is safe but pins this module below whatever the rest of objectionary targets, so `Phi` API added later at a higher level would not compile here and the reason would be non-obvious. If 11 is deliberate — the widest JVM this can be run on — a comment saying so is worth the line.

test/commands/test_inspect.js
```diff
@@ -0,0 +1,129 @@
...
+
+describe('inspect/java', () => {
+  const home = path.resolve('temp/test-inspect-unit');
+  let params;
+  let printed;
+  let killed;
+  before(async () => {
+    fs.rmSync(home, {recursive: true, force: true});
+    fs.mkdirSync(path.resolve(home, 'inspect'), {recursive: true});
+    fs.writeFileSync(path.resolve(home, 'inspect', 'inspect.jar'), '');
```
Favixx, line 64: Writing an empty file at the jar's path to stop `jar()` from building anything is a neat trick for keeping this suite fast, and it is also the thing that will keep the `existsSync` shortcut from ever being fixed: change `jar()` to rebuild on a stale source and this test starts invoking Maven, so whoever makes that change has to come back here and think about it.

Better to make the seam explicit — the module already takes `exec` and `runner` as optional parameters for exactly this reason; a third one for the jar builder, defaulting to the real `jar`, would let this test say `() => 'inspect.jar'` and stop depending on caching behaviour it does not actually want to assert.

test/commands/test_inspect.js
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Take a port the operating system is willing to give us right now, so that
+ * two suites on the same machine, or a JVM an earlier run left behind, do not
+ * collide on a number written into the test.
+ * @return {Promise<Number>} A port nobody listens on
+ */
+async function free() {
+  const socket = net.createServer();
+  await new Promise((resolve) => socket.listen(0, '127.0.0.1', resolve));
```
Favixx, line 22: The comment is right about the problem and the function does narrow it, but it is worth being honest that it does not close it: the port is released on line 24 and bound again by the JVM some milliseconds later, and anything on the machine can take it in between. On a busy CI runner that is a rare, unreproducible failure of the first test in this file, which is the worst kind to debug.

The way to actually close it is to have the server bind port 0 itself and report the number it got — which the `@todo` about ports on the JavaScript side will want anyway, since "a port that is already taken" stops being a problem the moment nobody chooses one.

test/commands/test_inspect.js
```diff
@@ -0,0 +1,129 @@
...
+        }
+      );
+    } finally {
+      console.info = info;
+      server.close();
+    }
+  });
+  it('runs the server class', () => {
+    assert(
+      params.includes('org.eolang.eoc.Inspect'),
```
Favixx, line 92: Three assertions on `params` — the class and both jars — and none on `--port`, which is the only argument the server treats as mandatory and the only one it will refuse to start without. A `params` array that had lost the flag, or passed the port before the class name, would satisfy every test in this file while `eoc inspect` failed for everybody.

`assert(params.includes('--port') && params[params.indexOf('--port') + 1] === String(server.address().port))` covers the contract that `Inspect.port()` depends on.

test/commands/test_inspect.js
```diff
@@ -0,0 +1,129 @@
...
+  it('prints the object the server answers with', () => {
+    assert(
+      printed.includes('@ Φ'),
+      `inspect does not print the object the session starts at: ${printed}`
+    );
+  });
+  it('kills the server when the session ends', () => {
+    assert(killed, 'inspect leaves the server running');
+  });
+  it('fails fast when javac is not on the PATH', async () => {
```
Favixx, line 117: The stub server on line 65 answers on the first request, so `ask()` is only ever exercised down its happy path: the retry, the 200ms wait, the deadline and the error message it throws are all untested. That is most of the function, and it is the part that decides what a user sees when the server does not come up.

Two tests would cover it without slowing the suite down: a stub that refuses the first two connections and answers the third, asserting the answer still arrives; and a call with a deadline already in the past against a port nobody listens on, asserting the thrown message names the port. The second one also pins the behaviour discussed on line 41 — today it would pass even if the server had answered with garbage.

src/eoc.js
```diff
@@ -318,6 +319,22 @@ program.command('dataize')
...
   });
 
+program.command('inspect')
+  .description('Traverse the tree of objects of a compiled program')
+  .option('--port <number>', 'TCP port for the inspection server', '8080')
+  .action(async (str, opts) => {
+    pin(program.opts());
+    clear(str);
+    const inspect = coms().inspect;
+    if (inspect === undefined) {
```
Favixx, line 329: This guard is good and I would rather it spread than shrink: `dataize` one screen above does the same `coms().dataize` lookup with no check, so running it under `--language=js` fails with `coms(...).dataize is not a function` instead of a sentence. Worth a follow-up puzzle rather than anything in this PR.

One wrinkle here: `coms()` is read twice, once through `coms().inspect` and once inside `pipe()(coms(), ...)`. Harmless, but the local `inspect` shadows the `require`d module name in this file's mental model — `command` or `run` would read better next to `pipe()`.

src/eoc.js
```diff
@@ -318,6 +319,22 @@ program.command('dataize')
...
+      throw new Error(`The "inspect" command only works for ${language.java}`);
+    }
+    if (program.opts().alone === undefined) {
+      await pipe()(coms(), ['register', 'assemble', 'lint', 'resolve', 'transpile', 'compile', 'link'], program.opts());
```
Favixx, line 333: With `--alone` this branch is skipped, so `link` never runs, and `inspect()` then puts `target/eoc.jar` on the classpath regardless of whether it exists. The JVM starts, fails to find the program, and the user gets a `NoClassDefFoundError` naming a class they never heard of.

That is the same shape as the problem the check four lines up solves so well. `fs.existsSync(path.resolve(opts.target, 'eoc.jar'))` inside `inspect.js`, throwing "the program is not linked yet, run without --alone or run `eoc link` first", turns the worst failure of this command into a sentence. `dataize` has the identical hole, which is an argument for putting the check in the command rather than here.

Favixx (changes requested):
The split is the right one. Putting the server in a jar of its own, leaving the EO jar alone, and letting the two sides talk over HTTP means the JavaScript side never has to know what an object is and the Java side never has to know what a terminal is — which is exactly the line the `@todo` at the top of `inspect.js` draws, in those words. `provided` scope on `eo-runtime` is right too, since the linked program brings its own. And the guard on line 329 that refuses the command outside Java is more care than `dataize` takes one screen up.

Three things I would like changed before this merges, and they are connected.

**The JSON is built by string concatenation.** `String.format("{\"forma\":\"%s\"}", Phi.Φ.forma())` produces valid JSON only while no `forma()` ever contains a quote or a backslash. The root's does not; the objects the four pending verbs will reach are user-named, and one of them will.

**A malformed answer is reported as a timeout.** `ask()` wraps the whole of `fetch(...).json()` in one `catch` that retries, so a server that answers instantly with a body that will not parse is retried for sixty seconds and then blamed for never opening the port. That is the message a user gets for the bug above.

**The jar is rebuilt only if it is missing.** `if (!fs.existsSync(built))` means that after the first `eoc inspect`, no change to `inspect/src` ever reaches the running server again — and `test_inspect.js` writes an empty file at that path precisely to exploit the behaviour, which fixes it in place.

The rest is smaller: an unvalidated `--port` on both sides, a `kill()` nobody waits for, the server's own output interleaved with the session's, and four gaps in the tests. Details inline. The puzzles are well written and I have no complaint about the scope — one question answered is a fine first slice.

morphqdd:
@yegor256, please check

morphqdd:
@yegor256, ping

morphqdd:
@yegor256, Could you please take a look at this? Would that get us started on working on the `inspect` command in `eoc`?

0crat:
@morphqdd Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (427 >= 100); -8 for way too many hits-of-code (427 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6707](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@thayorns Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+4460](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+56](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1163, opened 2026-08-03
Run the CLI through one helper in tests

### LINKED ISSUE #1124: Inconsistent functionality in test/helpers.js runSync leads to redundant implementations
## What Happens

`test/helpers.js` exposes `runSync` as the shared way to invoke the CLI, but it returns only stdout — its own JSDoc says `@return {String} Stdout`, and `execFileSync` forwards the child's stderr to the parent instead of capturing it. Any test that needs stderr or the exit status therefore cannot use it, and spawns `src/eoc.js` inline instead. That is now happening at seven sites in three files:

| File | Sites |
|---|---|
| `test/commands/test_clean.js` | 55, 69, 99, 113 |
| `test/test_eoc.js` | 55, 176 |
| `test/commands/test_transpile.js` | 60 |

`test/test_eoc.js` defines the identical helper twice in one file, at line 55 and line 176:

```js
const eoc = function(...args) {
  return spawnSync('node', [path.resolve('./src/eoc.js'), '--batch', ...args]);
};
```

The copies have already drifted from `runSync` and from each other:

1. **No timeout.** `runSync` passes `timeout: 1200000`. None of the seven inline spawns pass one, so a hung child blocks the run until CI kills the whole job.
2. **No `windowsHide`.** `runSync` passes `windowsHide: true`; none of the inline spawns do.
3. **Inconsistent `--batch`.** `runSync` always prepends it, and so do both copies in `test_eoc.js`, but all four sites in `test_clean.js` omit it — those tests run the CLI in interactive mode while every other test does not.

A flag added to `runSync` reaches none of the seven.

## What Should Happen

`test/helpers.js` should offer a sibling helper that performs one spawn and returns stdout, stderr, and status together, with the same `timeout` and `windowsHide` settings `runSync` already uses. The seven inline spawns should then be collapsed onto it, so child-process invocation is defined in exactly one place.

This is a test-only refactor: no product behavior changes.

Thayorns:
Tests invoked `src/eoc.js` in eight different places, each with its own `spawnSync` call. None of them passed a timeout, none passed `windowsHide`, and the four in `test_clean.js` ran the CLI without `--batch` while every other test used it. Two identical copies of the same private `eoc` helper sat in `test_eoc.js` alone.

`test/helpers.js` now has `runOutput`, a sibling of `runSync` that performs one spawn and hands back stdout, stderr and status together, with the timeout and `windowsHide` settings `runSync` already used. Every inline spawn in `test_clean.js`, `test_eoc.js`, `test_transpile.js` and `test_format.js` is collapsed onto it, so a flag added to the helper now reaches all of them. `test/test_helpers.js` covers the helper itself: stderr, exit status, stdout and the working directory it is given.

This is test-only, no product behaviour changes.

Closes #1124

### DIFF
--- a/test/commands/test_clean.js
+++ b/test/commands/test_clean.js
@@ -7,7 +7,7 @@ const assert = require('assert');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
-const {runSync} = require('../helpers');
+const {runSync, runOutput} = require('../helpers');
 
 describe('clean', () => {
   const testDir = 'temp/test-clean',
@@ -52,21 +52,16 @@ describe('clean', () => {
     done();
   });
   it('refuses to delete the current working directory', (done) => {
-    const {spawnSync} = require('child_process');
     const home = path.resolve(testDir, 'refuses-cwd');
     fs.rmSync(home, {recursive: true, force: true});
     fs.mkdirSync(home, {recursive: true});
-    const s = spawnSync(
-      'node', [path.resolve('./src/eoc.js'), '--target', '.', 'clean'],
-      {cwd: home}
-    );
-    assert(s.status !== 0, s.stderr.toString());
-    assert(s.stderr.toString().includes('Refusing to delete'), s.stderr.toString());
+    const outcome = runOutput(['--target', '.', 'clean'], {cwd: home});
+    assert(outcome.status !== 0, outcome.stderr);
+    assert(outcome.stderr.includes('Refusing to delete'), outcome.stderr);
     assert(fs.existsSync(home), 'the current directory must not be deleted');
     done();
   });
   it('refuses to delete the current directory through a symbolic link', (done) => {
-    const {spawnSync} = require('child_process');
     const home = path.resolve(testDir, 'refuses-symlink-cwd'),
       real = path.resolve(home, 'real'),
       link = path.resolve(home, 'link'),
@@ -74,20 +69,14 @@ describe('clean', () => {
     fs.rmSync(home, {recursive: true, force: true});
     fs.mkdirSync(real, {recursive: true});
     fs.symlinkSync(home, link, type);
-    const s = spawnSync(
-      'node',
-      [
-        path.resolve('./src/eoc.js'),
-        '--target',
-        path.resolve(link, 'real'),
-        'clean'
-      ],
+    const outcome = runOutput(
+      ['--target', path.resolve(link, 'real'), 'clean'],
       {cwd: real}
     );
-    assert(s.status !== 0, `${s.stdout.toString()}\n${s.stderr.toString()}`);
+    assert(outcome.status !== 0, `${outcome.stdout}\n${outcome.stderr}`);
     assert(
-      s.stderr.toString().includes('Refusing to delete'),
-      s.stderr.toString()
+      outcome.stderr.includes('Refusing to delete'),
+      outcome.stderr
     );
     assert(
       fs.existsSync(real),
@@ -96,29 +85,24 @@ describe('clean', () => {
     done();
   });
   it('refuses to delete an ancestor of the current working directory', (done) => {
-    const {spawnSync} = require('child_process');
     const home = path.resolve(testDir, 'refuses-ancestor'),
       nested = path.resolve(home, 'nested');
     fs.rmSync(home, {recursive: true, force: true});
     fs.mkdirSync(nested, {recursive: true});
-    const s = spawnSync(
-      'node', [path.resolve('./src/eoc.js'), '--target', '..', 'clean'],
-      {cwd: nested}
-    );
-    assert(s.status !== 0, s.stderr.toString());
+    const outcome = runOutput(['--target', '..', 'clean'], {cwd: nested});
+    assert(outcome.status !== 0, outcome.stderr);
     assert(fs.existsSync(home), 'the ancestor directory must not be deleted');
     done();
   });
   it('refuses to delete the home directory', (done) => {
-    const {spawnSync} = require('child_process');
     const fakeHome = path.resolve(testDir, 'fake-home');
     fs.rmSync(fakeHome, {recursive: true, force: true});
     fs.mkdirSync(fakeHome, {recursive: true});
-    const s = spawnSync(
-      'node', [path.resolve('./src/eoc.js'), '--target', fakeHome, 'clean'],
+    const outcome = runOutput(
+      ['--target', fakeHome, 'clean'],
       {env: {...process.env, HOME: fakeHome, USERPROFILE: fakeHome}}
     );
-    assert(s.status !== 0, s.stderr.toString());
+    assert(outcome.status !== 0, outcome.stderr);
     assert(fs.existsSync(fakeHome), 'the home directory must not be deleted');
     done();
   });
--- a/test/commands/test_format.js
+++ b/test/commands/test_format.js
@@ -6,8 +6,7 @@
 const assert = require('assert');
 const fs = require('fs');
 const path = require('path');
-const {spawnSync} = require('child_process');
-const {runSync, parserVersion, homeTag, weAreOnline} = require('../helpers');
+const {runSync, runOutput, parserVersion, homeTag, weAreOnline} = require('../helpers');
 
 const messy = '[] > app';
 const tidy = ['[] > app', ''].join('\n');
@@ -47,11 +46,7 @@ function args(home, sources) {
  * @return {Number} Exit code
  */
 function attempt(argv) {
-  return spawnSync(
-    'node',
-    [path.resolve('./src/eoc.js'), '--batch'].concat(argv),
-    {timeout: 1200000, windowsHide: true}
-  ).status;
+  return runOutput(argv).status;
 }
 
 const cases = [
--- a/test/commands/test_transpile.js
+++ b/test/commands/test_transpile.js
@@ -6,7 +6,9 @@
 const assert = require('assert');
 const fs = require('fs');
 const path = require('path');
-const {runSync, assertFilesExist, parserVersion, homeTag, weAreOnline} = require('../helpers');
+const {
+  runSync, runOutput, assertFilesExist, parserVersion, homeTag, weAreOnline
+} = require('../helpers');
 
 describe('transpile', () => {
   before(weAreOnline);
@@ -54,12 +56,9 @@ describe('transpile', () => {
     done();
   });
   it('attempts to transpile a simple .EO program to wrong platform', (done) => {
-    const {spawnSync} = require('child_process'),
-      s = spawnSync(
-        'node', [path.resolve('./src/eoc.js'), 'transpile', '--language=Eiffel']
-      );
-    assert(s.status != 0);
-    assert(s.stderr.includes('Unknown platform Eiffel'), s.stderr);
+    const outcome = runOutput(['transpile', '--language=Eiffel']);
+    assert(outcome.status != 0);
+    assert(outcome.stderr.includes('Unknown platform Eiffel'), outcome.stderr);
     done();
   });
 });
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -44,6 +44,31 @@ module.exports.runSync = function runSync(args) {
 };
 
 /**
+ * Helper to run EOC command line tool and capture everything it produces.
+ *
+ * @param {Array} args - Array of args
+ * @param {Object} options - Extra options for the child process
+ * @return {Object} Stdout, stderr and exit status
+ */
+module.exports.runOutput = function runOutput(args, options = {}) {
+  const {spawnSync} = require('child_process');
+  const outcome = spawnSync(
+    'node',
+    [path.resolve('./src/eoc.js'), '--batch'].concat(args),
+    {
+      'timeout': 1200000,
+      'windowsHide': true,
+      ...options,
+    }
+  );
+  return {
+    'stdout': outcome.stdout.toString(),
+    'stderr': outcome.stderr.toString(),
+    'status': outcome.status,
+  };
+};
+
+/**
  * Assert that all files exist.
  *
  * @param {String} stdout - The stdout printed
--- a/test/test_eoc.js
+++ b/test/test_eoc.js
@@ -5,7 +5,7 @@
 
 const assert = require('assert');
 const version = require('../src/version');
-const {runSync, jeoVersion, weAreOnline} = require('./helpers');
+const {runSync, runOutput, jeoVersion, weAreOnline} = require('./helpers');
 
 describe('eoc', () => {
   it('prints its own version', (done) => {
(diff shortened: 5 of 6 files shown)


Thayorns:
@yegor256 take a look please, happy to clarify anything here

test/test_helpers.js
```diff
@@ -0,0 +1,38 @@
...
+  });
+  it('captures the stdout of a successful command', () => {
+    const outcome = runOutput(['--version']);
+    assert(
+      outcome.stdout.trim().length > 0,
+      `stdout of a successful command is empty: ${outcome.stdout}`
+    );
+  });
+  it('runs the command in the given directory', () => {
+    const outcome = runOutput(['--target', '.', 'clean'], {cwd: require('os').tmpdir()});
```
Favixx, line 32: This is the one I would not merge. `os.tmpdir()` is a real directory with real contents — other suites' scratch files, other processes' sockets — and this runs `clean --target .` inside it, relying entirely on the CLI refusing.

When the guard works, the test passes. When it breaks, the test does not report a failure; it destroys the temp directory first and reports afterwards, on every machine that runs the suite, including the developer's. A test for a safety net must not be the thing the net is catching.

`fs.mkdtempSync(path.join(os.tmpdir(), 'eoc-helpers-'))` gives a directory that is real, empty and disposable, and the assertion is unchanged. `test_eoc.js` already uses that idiom a few lines from here.

Smaller point on the same line: `require('os')` inline inside the argument. This PR's whole subject is that inline requires in tests are worth hoisting — `os` belongs at the top of the file with `assert`.

test/helpers.js
```diff
@@ -43,6 +43,31 @@ module.exports.runSync = function runSync(args) {
...
+    {
+      'timeout': 1200000,
+      'windowsHide': true,
+      ...options,
+    }
+  );
+  return {
+    'stdout': outcome.stdout.toString(),
+    'stderr': outcome.stderr.toString(),
+    'status': outcome.status,
```
Favixx, line 67: `outcome.error` is dropped, and `status` is passed through as-is — which is `null` whenever the child was killed rather than exited, the timeout on line 59 being the case that matters here.

Every negative assertion this PR touches is `status !== 0` or `notStrictEqual(status, 0)`, and both are satisfied by `null`. So a `clean` that hung for twenty minutes and got killed makes `refuses to delete the current working directory` pass — after the second assertion on stderr, which would have been empty, so in practice it fails there. But `refuses to delete an ancestor of the current working directory` on line 92 asserts only the status and the directory still existing, and a killed CLI satisfies both.

Returning `error` in the triple, or normalising a killed run to a status that cannot be confused with a clean failure, closes it. Given that the absence of timeouts is one of the four problems this PR names, it would be a shame for the timeout to become invisible instead.

Favixx (commented):
Good change and the description is accurate about what was wrong: eight hand-rolled `spawnSync` calls, none of them with a timeout, two copies of the same private `eoc` closure in one file, and four tests silently running the CLI in a different mode from every other test. Collapsing them onto one helper is the right fix, and `runOutput` returning the triple rather than the raw `SpawnSyncReturns` is what makes the call sites read as well as they now do — `assert(outcome.status !== 0, outcome.stderr)` against the old two-line `toString()` dance is a real gain.

Two things I would change before this merges.

**`test_helpers.js` runs `clean --target .` with the working directory set to the system temp directory.** It asserts that the CLI refuses. If the guard it depends on ever regresses, that test does not fail — it deletes the contents of `/tmp`, on every developer machine and every CI runner, before anyone sees a red build. A test whose failure mode is destructive should not point at a real directory; a freshly made temporary one costs two lines.

**`runOutput` throws away `outcome.error` and passes `status` through unchanged.** A run that hits the twenty-minute timeout comes back with `status: null`, and `null !== 0` is true, so every negative assertion in this PR — six of them — treats a timed-out CLI as a correctly failing one. That is the exact class of bug the missing timeouts were causing before, moved one level up.

The rest is small: an inline `require` in the helper the PR exists to de-inline, the same two options still duplicated between `runSync` and `runOutput`, and a helper test suite that tests the CLI rather than the helper.

test/helpers.js
```diff
@@ -43,6 +43,31 @@ module.exports.runSync = function runSync(args) {
...
+    'node',
+    [path.resolve('./src/eoc.js'), '--batch'].concat(args),
+    {
+      'timeout': 1200000,
+      'windowsHide': true,
+      ...options,
+    }
+  );
+  return {
+    'stdout': outcome.stdout.toString(),
```
Favixx, line 65: `spawnSync` sets `stdout` and `stderr` to `null`, not to an empty buffer, when the child could not be spawned at all — `node` missing from `PATH`, or `ENOMEM` under a loaded CI runner. Both `toString()` calls then throw `TypeError: Cannot read properties of null`, from inside a helper, with a stack pointing at `helpers.js` and nothing about which command was being run.

That is a worse failure than the one it replaces: the old inline `s.stderr.toString()` had the same problem, but it was in the test that owned it. Now one helper turns every such failure into the same opaque error.

`(outcome.stdout || '').toString()` is enough, or check `outcome.error` first and throw something that names `args`.

test/helpers.js
```diff
@@ -43,6 +43,31 @@ module.exports.runSync = function runSync(args) {
...
 
+/**
+ * Helper to run EOC command line tool and capture everything it produces.
+ *
+ * @param {Array} args - Array of args
+ * @param {Object} options - Extra options for the child process
+ * @return {Object} Stdout, stderr and exit status
+ */
+module.exports.runOutput = function runOutput(args, options = {}) {
+  const {spawnSync} = require('child_process');
```
Favixx, line 54: An inline `require` inside the function, in the helper whose reason for existing is that the tests had inline requires. `path` is already imported at the top of this file and `runOutput` uses that one on the next line, so the asymmetry is visible in two adjacent lines.

`runSync` above does something slightly worse — it re-requires `path` locally, shadowing the module-level import — so if you are moving this one up, that shadow is worth removing in the same commit.

test/helpers.js
```diff
@@ -43,6 +43,31 @@ module.exports.runSync = function runSync(args) {
...
+  const outcome = spawnSync(
+    'node',
+    [path.resolve('./src/eoc.js'), '--batch'].concat(args),
+    {
+      'timeout': 1200000,
```
Favixx, line 59: `timeout` and `windowsHide` are now written twice in this file, once here and once in `runSync` twenty lines up, with the same values. The description promises that "a flag added to the helper now reaches all of them", and that is true of the eight call sites but not of the two helpers: adding a third setting still means editing two places, and the two silently drifting apart is exactly the state this PR is cleaning up.

Since `runOutput` is the more general of the two — `runSync` is it, plus "throw when the status is not zero, and print stdout first" — the natural shape is `runSync` calling `runOutput`. That also gives `runSync` the `options` parameter it does not have, which is what forced four `test_clean.js` tests to hand-roll a spawn in the first place, because they needed `cwd` and `env`.

test/helpers.js
```diff
@@ -43,6 +43,31 @@ module.exports.runSync = function runSync(args) {
...
+ *
+ * @param {Array} args - Array of args
+ * @param {Object} options - Extra options for the child process
+ * @return {Object} Stdout, stderr and exit status
+ */
+module.exports.runOutput = function runOutput(args, options = {}) {
+  const {spawnSync} = require('child_process');
+  const outcome = spawnSync(
+    'node',
+    [path.resolve('./src/eoc.js'), '--batch'].concat(args),
```
Favixx, line 57: `--batch` is unconditional here, which is the change that makes the four `test_clean.js` tests run in a mode they did not run in before. I think it is the right default and the description is right to call the old inconsistency a defect — but it is a behavioural change to existing tests, not a pure refactor, and the JSDoc above says nothing about it.

Worth one line in the doc saying that the CLI is always run in batch mode, so that the next person who needs to test interactive behaviour knows to add a parameter rather than assuming the helper is neutral. Right now the only way to find out is to read the implementation.

test/helpers.js
```diff
@@ -43,6 +43,31 @@ module.exports.runSync = function runSync(args) {
...
+    {
+      'timeout': 1200000,
+      'windowsHide': true,
+      ...options,
```
Favixx, line 61: Spreading `options` after the defaults is the right order — a caller who needs a shorter timeout can have one — but it also means a caller can pass `stdio` or `encoding` and quietly change what the two `toString()` calls below produce, with no error and no hint.

The four callers today pass only `cwd` and `env`, which is exactly what the parameter is for. A sentence in the JSDoc naming those two as the supported keys would keep it that way, and is cheaper than defending against the rest.

test/test_helpers.js
```diff
@@ -0,0 +1,38 @@
...
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+
+const assert = require('assert');
+const {runOutput} = require('./helpers');
+
+describe('helpers', () => {
+  it('captures the stderr of a failing command', () => {
+    const outcome = runOutput(['transpile', '--language=Eiffel']);
```
Favixx, line 11: This is `test_transpile.js` line 59 verbatim — same command, same expected message. So the CLI's wording for an unknown platform is now asserted in two files, and changing it breaks a test whose name says it is about capturing stderr.

A test for the helper wants a command whose stderr is stable and uninteresting, so that the test fails only when the capture breaks. Since the point is only that stderr is not lost, `runOutput(['--no-such-flag'])` and asserting that stderr is non-empty says the same thing while depending on nothing that anyone will reword.

test/test_helpers.js
```diff
@@ -0,0 +1,38 @@
...
+    assert(
+      outcome.stderr.includes('Unknown platform Eiffel'),
+      `stderr of a failing command is lost: ${outcome.stderr}`
+    );
+  });
+  it('reports the exit code of a failing command', () => {
+    const outcome = runOutput(['transpile', '--language=Cobol']);
```
Favixx, line 18: `Cobol` here and `Eiffel` one test up, running the same command through the same code path — the only difference between these two tests is which assertion they make about the same outcome. One `runOutput` call and two assertions would cost one process instead of two, and each of these spawns a full Node process.

Worth noting what neither of them checks: that `status` is the number the CLI actually exited with. `notStrictEqual(status, 0)` is satisfied by `null`, which is what a killed child gives — the case discussed on `helpers.js` line 67. Asserting `strictEqual(outcome.status, 1)` would make this test the one that catches it.

test/test_helpers.js
```diff
@@ -0,0 +1,38 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+
+const assert = require('assert');
+const {runOutput} = require('./helpers');
+
+describe('helpers', () => {
```
Favixx, line 9: Four tests, and between them they cover stdout, stderr, status and `cwd` — but none of the three things this helper was actually created to guarantee: that a timeout is applied, that `windowsHide` is set, and that `--batch` is prepended. Those are the reasons the description gives for the helper existing, and they are the part that silently differed across the eight call sites before.

The last one is testable cheaply and is the one with a behavioural consequence for `test_clean.js`: a command that reports its own flags, or asserting that `runOutput` and a bare `spawnSync` of the same arguments disagree, pins it. The timeout is testable by passing `{timeout: 1}` through `options` and asserting the run comes back killed — which would also have surfaced the `status: null` question.

test/commands/test_clean.js
```diff
@@ -96,29 +85,24 @@ describe('clean', () => {
...
     const home = path.resolve(testDir, 'refuses-ancestor'),
       nested = path.resolve(home, 'nested');
     fs.rmSync(home, {recursive: true, force: true});
     fs.mkdirSync(nested, {recursive: true});
-    const s = spawnSync(
-      'node', [path.resolve('./src/eoc.js'), '--target', '..', 'clean'],
-      {cwd: nested}
-    );
-    assert(s.status !== 0, s.stderr.toString());
+    const outcome = runOutput(['--target', '..', 'clean'], {cwd: nested});
```
Favixx, line 92: This one asserts only that the status is not zero and that the directory survives, and both hold for a CLI that was killed by the helper's timeout without doing anything at all. The three siblings around it also assert on stderr, which makes them robust to that; this one does not.

One `assert(outcome.stderr.includes('Refusing to delete'), outcome.stderr)` here would bring it in line with the other three and remove the ambiguity. The message is the same one they check.

test/commands/test_clean.js
```diff
@@ -52,42 +52,31 @@ describe('clean', () => {
...
     const home = path.resolve(testDir, 'refuses-cwd');
     fs.rmSync(home, {recursive: true, force: true});
     fs.mkdirSync(home, {recursive: true});
-    const s = spawnSync(
-      'node', [path.resolve('./src/eoc.js'), '--target', '.', 'clean'],
-      {cwd: home}
-    );
-    assert(s.status !== 0, s.stderr.toString());
-    assert(s.stderr.toString().includes('Refusing to delete'), s.stderr.toString());
+    const outcome = runOutput(['--target', '.', 'clean'], {cwd: home});
```
Favixx, line 58: Worth calling out in the description that this test — and the three below it — now run with `--batch`, where before they did not. The PR body mentions the inconsistency, but reads as though adding `--batch` were a tidy-up rather than a change of what is being exercised.

It is almost certainly harmless for `clean`, whose guard runs before anything interactive. But these four are the safety tests for a destructive command, so "almost certainly" is worth converting into a sentence saying that batch mode does not affect the refusal path.

test/commands/test_format.js
```diff
@@ -47,11 +46,7 @@ function args(home, sources) {
  * @return {Number} Exit code
  */
 function attempt(argv) {
-  return spawnSync(
-    'node',
-    [path.resolve('./src/eoc.js'), '--batch'].concat(argv),
-    {timeout: 1200000, windowsHide: true}
-  ).status;
+  return runOutput(argv).status;
```
Favixx, line 49: The body is now a one-liner that forwards to the helper, which is good, but the JSDoc above still promises `@return {Number} Exit code` and `runOutput` can hand back `null` for a killed child.

`attempt` is used by the table of cases below, so a `null` there compares unequal to whatever exit code each case expects and the failure message will say `expected 1, got null` — which is at least honest. Still, the doc should say `{Number|null}`, or the helper should stop producing `null`, which I would prefer.

test/commands/test_transpile.js
```diff
@@ -54,12 +56,9 @@ describe('transpile', () => {
...
   });
   it('attempts to transpile a simple .EO program to wrong platform', (done) => {
-    const {spawnSync} = require('child_process'),
-      s = spawnSync(
-        'node', [path.resolve('./src/eoc.js'), 'transpile', '--language=Eiffel']
-      );
-    assert(s.status != 0);
-    assert(s.stderr.includes('Unknown platform Eiffel'), s.stderr);
+    const outcome = runOutput(['transpile', '--language=Eiffel']);
+    assert(outcome.status != 0);
```
Favixx, line 60: Neutral note, since it took me a second look: the old code asserted `s.stderr.includes(...)` on a `Buffer`, which works because `Buffer.prototype.includes` accepts a string, so this is behaviour-preserving rather than a fix. Nothing lost.

While this line is being touched, `!=` could become `!==` — the file uses strict comparison everywhere else, and eslint's `eqeqeq` would have caught it if it were on.

test/test_eoc.js
```diff
@@ -52,26 +52,22 @@ describe('eoc', () => {
 });
 
 describe('eoc', () => {
-  const {spawnSync} = require('child_process'),
-    path = require('path'),
-    eoc = function(...args) {
-      return spawnSync('node', [path.resolve('./src/eoc.js'), '--batch', ...args]);
-    };
+  const path = require('path');
```
Favixx, line 55: `const path = require('path')` survives inside the `describe`, now as the only thing left of the block that used to hold the private `eoc` closure. It is used once, forty lines below, for `path.resolve('./src/eoc.js')` in the `generate_comments` case.

Since the point of this PR is that requires belong at the top of the file, this is the last one worth moving — and the second `describe` block further down has already had its `path` and `child_process` requires removed entirely, so the file is currently inconsistent with itself.

Thayorns:
@yegor256 ping

Thayorns:
@yegor256 take a look

Thayorns:
@yegor256 take a look at this one please

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (176 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3804](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@favixx Thanks for the review! You've earned +12 points.  Your running score is [+68](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1212, opened 2026-08-31
#1173: drop the main field that pointed at nothing

### LINKED ISSUE #1173: `main` field in `package.json` incorrectly pointing to non-existent `eoc` file, causing require failures
## What happens

`package.json` declares `"main": "eoc"`, but there is no `eoc.js` or `eoc/` directory at the repo root - the actual entry point lives at `src/eoc.js` and is correctly declared under `"bin"`. Requiring the package as a library fails:

```
$ node -e "require('.')"
Error: Cannot find module '.../eoc'. Please verify that the package.json has a
valid "main" entry
```

This is harmless for normal CLI usage (`npx eolang` / the installed `eoc` binary both go through `bin`, not `main`), but breaks for anything that does `require('eolang')` from another Node package.

## What should happen

`"main"` should either point at the real entry file (`src/eoc.js`) or be removed if the package isn't meant to be required as a library.

morphqdd:
`"main": "eoc"` named a file that does not exist, so `require('eolang')` failed with `Cannot find module`.

Pointing it at `src/eoc.js` would not help: that file calls `program.parse(process.argv)` when it loads, so requiring the package would run the CLI. This is a command line tool and its entry point is already declared under `bin`, so the field is removed.

Closes #1173

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "oop"
   ],
   "license": "MIT",
-  "main": "eoc",
   "name": "eolang",
   "nyc": {
     "all": true,


yegor256:

---

zaguzovmaksim0-hue (changes requested):
The premise of this fix is no longer true on current `master`, so I don't think deleting `main` is the best resolution now.

`src/eoc.js` currently wraps CLI execution in `if (require.main === module)`, so requiring that file does **not** call `program.parseAsync(process.argv)`. It also deliberately exports `commandsDescription`, `canonicalLanguage`, and `select`, and the test suite imports this module directly. In other words, the entry file is now safe to load and already has a small programmatic surface.

With this PR, `require('eolang')` still fails — Node just falls back to a missing `index.js`. Changing `main` from `eoc` to `src/eoc.js` would instead fix the reported package-level import while preserving the CLI under `bin`.

Please update the metadata to point `main` at the guarded real entry file (and ideally add a regression that requires the package root) rather than removing the field.

PR #1213, opened 2026-08-31
#1166: default print to a directory the pipeline makes

### LINKED ISSUE #1166: `eoc print` default `--print-input 2-optimize` points at a directory the pipeline never creates
## What happens

`eoc print` defaults `--print-input` to `2-optimize` (`src/eoc.js:226-230`, consumed at `src/commands/print.js:17`). The current eo-maven-plugin pipeline never creates a `2-optimize` directory: a full `transpile --track-transformation-steps` run produces `1-parse`, `3-lint`, `4-resolve`, `5-pre-transpile`, `5-transpile`, `classes`, `eo`, `generated-sources` - never `2-optimize`.

Since `eo:print`'s `printSourcesDir` points at a directory that does not exist, it is a silent no-op, and `print.js` still prints a success message afterward regardless:

```
$ node src/eoc.js --batch --easy -s src -t target print
XMIR files converted to EO files at ../target/print in 5s
$ ls target/print
ls: cannot access 'target/print': No such file or directory

$ node src/eoc.js --batch --easy -s src -t target print --print-input 1-parse
XMIR files converted to EO files at ../target/print in 8s
$ find target/print
target/print
target/print/simple.eo
```

So the command itself works correctly once pointed at a directory that actually exists - only the default value is stale. All three cases in `test/commands/test_print.js` pass an explicit `--print-input=input`, which is why the default has never been exercised by a test.

## What should happen

`eoc print` run with no `--print-input` should convert something, the way it does when pointed at a real stage directory, instead of silently producing zero output while still reporting success. The default should be updated to a directory the current pipeline actually produces (e.g. `1-parse`), and it would help to add a test that exercises the default value specifically, since none currently does.

morphqdd:
The default pointed at `2-optimize`, which the current pipeline never creates: a tracked run leaves `1-parse`, `3-lint`, `4-resolve`, `5-pre-transpile` and `5-transpile`. `eo:print` was therefore given a directory that does not exist, printed nothing, and `print.js` reported success anyway.

`1-parse` is the stage `eoc print`'s own `--alone` path already names a few hundred lines further down in the same file.

Closes #1166

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -227,7 +227,7 @@ program.command('print')
   .option(
     '--print-input <dir>',
     'Directory where XMIR files for translation to EO are taken (relative to --target)',
-    '2-optimize'
+    '1-parse'
   )
   .option(
     '--print-output <dir>',


yegor256:

---

Issue #424, opened 2024-12-15
semver failure when using `1.0-SNAPSHOT`


yegor256:
I'm getting this when I use `--parser=1.0-SNAPSHOT`:

```
/opt/homebrew/lib/node_modules/eolang/node_modules/semver/classes/semver.js:38
      throw new TypeError(`Invalid Version: ${version}`)
            ^

TypeError: Invalid Version: 1.0-SNAPSHOT
    at new SemVer (/opt/homebrew/lib/node_modules/eolang/node_modules/semver/classes/semver.js:38:13)
    at compare (/opt/homebrew/lib/node_modules/eolang/node_modules/semver/functions/compare.js:3:3)
    at Object.gte (/opt/homebrew/lib/node_modules/eolang/node_modules/semver/functions/gte.js:2:30)
    at module.exports [as lint] (/opt/homebrew/lib/node_modules/eolang/src/commands/lint.js:39:14)
    at /opt/homebrew/lib/node_modules/eolang/src/eoc.js:341:29
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

yegor256:
@rultor release, tag is `0.27.1`

PR #1214, opened 2026-08-31
#1170: name the option when the parser version is malformed

### LINKED ISSUE #1170: `--parser` with a malformed version prints a bare `Invalid Version` message with no option context
## What happens

`src/commands/lint.js:64` calls `semver.gte(opts.parser, '0.45.0')` inside `goals()`. For a malformed, non-SNAPSHOT `--parser` value, `semver.gte` throws a raw error that reaches the user as a bare one-line message with no indication of which option caused it:

```
$ node src/eoc.js --parser 1.0 lint ; echo rc=$?
Invalid Version: 1.0
rc=1

$ node src/eoc.js --parser latest lint
Invalid Version: latest
```

Compare this to the `--language` handling added in #1065, which names the offending option clearly: `error: option '-l, --language <name>' argument 'x' is invalid. Unknown platform x`.

`goals()` runs in `java/pipeline.js:31`, before `flags()` on line 32, so `mvnw.js`'s own Maven Central version-check guard never gets a chance to produce its more helpful message either.

This is related to but distinct from the closed #424 ("semver failure when using `1.0-SNAPSHOT`"), which fixed only the `-SNAPSHOT` short-circuit case - a plain malformed version like `1.0` or `latest` was never covered.

## What should happen

A malformed `--parser` value should produce a message naming the option, the same way `--language` does, rather than a bare `Invalid Version: X` with no context about which flag was wrong.

morphqdd:
`semver.gte` was reached inside `goals()` and threw a bare `Invalid Version: 1.0` with nothing saying which flag was wrong. `goals()` runs before `flags()`, so `mvnw.js`'s own guard never got a chance either.

The value is now checked where commander parses it, so the message names the option, the same way `--language` does:

```
error: option '--parser <version>' argument '1.0' is invalid. 1.0 is not a version
```

The `-SNAPSHOT` short-circuit that #424 added is kept, since `1.0-SNAPSHOT` is not valid semver and is a value people actually pass.

Closes #1170

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -121,6 +121,7 @@ if (process.argv.includes('--latest')) {
   console.debug(`EO parser ${parser}; use the --latest flag if you need a fresher one`);
 }
 
+const semver = require('semver');
 const version = require('./version');
 program
   .name('eoc')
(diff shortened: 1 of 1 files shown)


morphqdd:

yegor256:

---

PR #1223, opened 2026-08-31
#1221: kill the inspection server on a signal too

### LINKED ISSUE #1221: JVM gets abandoned in inspect.js resulting in execution failure
The puzzle `500-0f2b7f36` from #500 has to be resolved: 

https://github.com/objectionary/eoc/blob/d83dd24378e80ad5a60d9a558467a81b7193eed7/src/commands/java/inspect.js#L67-L72

The puzzle was created by morphqdd on 31-Aug-26. 

morphqdd:
The server was killed in the `finally`, which covers a failure of our own code but not a Ctrl-C and not a crash of the Node process. In both cases the JVM stayed alive holding the port, and the next run of the command could not bind it.

The child is now killed on `SIGINT`, `SIGTERM` and on `exit`, and the handlers are removed when the session ends normally. When the JVM exits before the port is open, which is what a port taken by something else looks like from here, the error names the port instead of waiting a minute for a timeout.

Closes #1221

### DIFF
--- a/src/commands/java/inspect.js
+++ b/src/commands/java/inspect.js
@@ -32,13 +32,20 @@ async function jar(opts) {
  * "localhost" resolves to ::1 first on some machines and is refused there.
  * @param {Number} port - TCP port the server listens on
  * @param {Number} deadline - Moment in time after which we give up, in millis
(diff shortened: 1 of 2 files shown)


yegor256:

---

PR #1224, opened 2026-08-31
#1215: load every object of the program at startup

### LINKED ISSUE #1215: Complete object tree not loading in Inspect.java
The puzzle `500-d4dc48ed` from #500 has to be resolved: 

https://github.com/objectionary/eoc/blob/d83dd24378e80ad5a60d9a558467a81b7193eed7/inspect/src/main/java/org/eolang/eoc/Inspect.java#L28-L35

The puzzle was created by @morphqdd on 31-Aug-26. 

Estimate: 60 minutes,  role: DEV. 

morphqdd:
A package object learns that a child exists only when that child is taken by name, so nothing could be listed under the root: the server could reach one object, lazily.

`Universe` reads the names once from the jars on the classpath and takes every one of them, which is the only moment the classpath is looked at. From then on the objects answer for themselves, which is what the verbs still to come need. The count goes into the answer and the JavaScript side prints the `Loaded N objects` line the session in #500 shows.

An object that refuses to be built is left out instead of stopping the load, since this tool exists for programs that do not run.

Closes #1215

### DIFF
--- a/inspect/src/main/java/org/eolang/eoc/Inspect.java
+++ b/inspect/src/main/java/org/eolang/eoc/Inspect.java
@@ -8,6 +8,7 @@ package org.eolang.eoc;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import org.eolang.Phi;
 import org.takes.Take;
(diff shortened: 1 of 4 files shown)


yegor256:

---

PR #1225, opened 2026-08-31
#1220: turn the single question into an interactive session

### LINKED ISSUE #1220: Single question in inspect.js needs rewriting for clarity
The puzzle `500-3dbc9ac9` from #500 has to be resolved: 

https://github.com/objectionary/eoc/blob/d83dd24378e80ad5a60d9a558467a81b7193eed7/src/commands/java/inspect.js#L60-L66

The puzzle was created by morphqdd on 31-Aug-26. 

Estimate: 60 minutes,  role: DEV. 

morphqdd:
The command asked one question and shut the server down. Now a prompt stays open, reads a verb, sends it to the server and prints the answer, until the user types `quit` or closes the terminal.

The verbs themselves are still refused, by name, which is what the four puzzles left in `Inspect` are about. What lands here is the loop, the prompt and the transport, with the printing kept on this side, since the server is the only one that knows the objects and this side is the only one that knows the terminal.

The prompt only opens on a real terminal, so the integration test, which runs the command with no tty, still ends after the first answer.

Built on #1223's branch, which is what keeps the JVM from outliving a session that now stays open. Once that is merged the diff here is just this change.

Closes #1220

### DIFF
--- a/inspect/src/main/java/org/eolang/eoc/Inspect.java
+++ b/inspect/src/main/java/org/eolang/eoc/Inspect.java
@@ -14,6 +14,7 @@ import org.takes.facets.fork.FkRegex;
 import org.takes.facets.fork.TkFork;
 import org.takes.http.Exit;
 import org.takes.http.FtBasic;
+import org.takes.rq.RqHref;
 import org.takes.rs.RsText;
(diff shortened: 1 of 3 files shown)


yegor256:

---

PR #1226, opened 2026-08-31
#1172: give each version-up workflow a branch of its own

### LINKED ISSUE #1172: `up.yml` and `eo-version-up.yml` workflows overwrite each other's changes due to shared `version-up` branch
## What happens

`.github/workflows/up.yml:28` and `.github/workflows/eo-version-up.yml:28` both open a pull request against the same branch name (`branch: version-up`), both set `delete-branch: true`, and both are triggered by the same event (`on: push: branches: [master], tags: ['*']`).

`peter-evans/create-pull-request` force-updates the named branch to base plus its own changes. Two jobs racing on the same push, targeting the same branch name, will overwrite each other's content - one bumps the README version reference, the other bumps `eo-version.txt` - and `delete-branch: true` on whichever PR merges first removes the branch out from under the other, unrelated job.

This is read-only evidence from the workflow files (not reproducible locally, since it needs GitHub Actions to actually race), so confidence is lower than the other findings in this batch, but the branch-name collision itself is plain from reading the two files side by side.

## What should happen

The two workflows should use distinct branch names (e.g. `version-up-readme` and `version-up-eo`), so a shared push trigger cannot make one job's PR clobber or delete the other's.

morphqdd:
Both workflows opened a pull request on `version-up`, both with `delete-branch: true`, and both fire on the same push. `peter-evans/create-pull-request` force-updates the named branch to base plus its own changes, so one job's content overwrites the other's, and whichever PR merges first deletes the branch out from under the other.

One branch each: `version-up-readme` for the README bump, `version-up-eo` for `eo-version.txt`.

Closes #1172

### DIFF
--- a/.github/workflows/eo-version-up.yml
+++ b/.github/workflows/eo-version-up.yml
@@ -25,7 +25,7 @@ jobs:
       - uses: peter-evans/create-pull-request@v8
         with:
           sign-commits: true
-          branch: version-up
+          branch: version-up-eo
           commit-message: 'new version of eo maven plugin'
           delete-branch: true
           title: 'New version of EO Maven Plugin'
--- a/.github/workflows/up.yml
+++ b/.github/workflows/up.yml
@@ -25,7 +25,7 @@ jobs:
       - uses: peter-evans/create-pull-request@v8
         with:
           sign-commits: true
-          branch: version-up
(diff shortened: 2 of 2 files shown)


yegor256:

---

Issue #982, opened 2026-06-06
`itest` build broken on master: `eo-version.txt` and `home-tag.txt` out of sync


Thayorns:
The build is broken: the `itest` workflow fails on `master`. The latest run is red with 25 `token recognition error at: '/'` errors across the pulled objects (`number.eo`, `bytes.eo`, `malloc.eo`, `sm/posix.eo` and others), `eo:transpile` then aborts with `Expected 1 child nodes, but found 0`, and Maven exits 1. Runs on 2026-06-04 and earlier were green.

It's probably a version mismatch. `eo-version.txt:1` pins the parser plugin to `0.61.0`, while `home-tag.txt:1` pins the pulled objects to `0.59.7`. Those objects use the typed-tail syntax, for example `[] > as-i64 /i64` in `itest/.eoc/2-pull/number.eo:29`, and the lexer in `0.61.0` does not recognize the `/`. `eo-version.txt` was bumped to `0.61.0` in commit 9031991 while `home-tag.txt` stayed at `0.59.7`.

Solution: align the two versions. Either bump `home-tag.txt` to an objects tag whose syntax `0.61.0` can parse, or pin `eo-version.txt` back to a plugin that understands the `0.59.7` objects.

Thayorns:
@yegor256 take a look

Thayorns:
@yegor256 have a look

PR #1227, opened 2026-08-31
#1171: bump home-tag together with the plugin version

### LINKED ISSUE #1171: `eo-version-up` workflow neglects to update `home-tag.txt`, causing build failures similar to #982
## What happens

`.github/workflows/eo-version-up.yml` (lines 21-24) bumps only `eo-version.txt` when a new `objectionary/eo` release comes out. `home-tag.txt` is never touched by any workflow (`grep -rn "home-tag" .github/` finds nothing).

Git history confirms the automated bump commits touch a single file:

```
$ git log --oneline -12 -- eo-version.txt
bacf6a2 new version of eo maven plugin
1c20008 new version of eo maven plugin (#975)
9031991 new version of eo maven plugin
...
$ git log --oneline -12 -- home-tag.txt
(none of the above commits appear here)
```

This is exactly the situation issue #982 diagnosed: "`eo-version.txt` was bumped to `0.61.0` in commit `9031991` while `home-tag.txt` stayed at `0.59.7`", which broke the `itest` build. That issue was closed by manually realigning the two files by hand, but the workflow that caused the drift was never changed, and there is still no CI guard asserting the two files agree.

## What should happen

Either `eo-version-up.yml` should bump `home-tag.txt` alongside `eo-version.txt` in the same commit, or a CI check should fail the build when the two files disagree. As it stands, the next automatic version bump will reintroduce the exact failure #982 already diagnosed once.

morphqdd:
The workflow bumped only `eo-version.txt`, and nothing anywhere touched `home-tag.txt`, so the two drifted apart on every release. That is exactly what #982 diagnosed: the files were realigned by hand and the workflow that pulled them apart was left as it was, so the next automatic bump would have done it again.

The tag is now written to both files in the same step, so they move together in one commit.

Closes #1171


yegor256:
This looks like the fix for the home-tag half of #1233, but build/itest are still red on the PR itself. Not merging until CI is green — please check whether the merge-pipeline half of #1233 is also needed for this to go green.

---

morphqdd:
CI is green now, on all three platforms for both `build` and `itest`.

On the other half of #1233: yes, `eo:merge` is still needed, but it belongs in a separate pull request. This one only keeps `home-tag.txt` in step with the plugin version, which is what makes the pulled objects come from the tag instead of from home's master. Adding the goal to the pipeline touches different files and can be reviewed on its own.

@yegor256, please take another look.

PR #1228, opened 2026-08-31
#1123: pass the program its arguments unchanged

### LINKED ISSUE #1123: `dataize`'s `js` backend removes `-`-prefixed arguments with no warning
## What happens

`eoc dataize`'s `js` platform backend (`src/commands/js/dataize.js:17`) silently drops every program argument that starts with `-`:

```js
module.exports = function(obj, args, opts) {
  return eo2jsw(
    ['dataize', obj, ...args.filter((arg) => !arg.startsWith('-'))].join(' '),
  );
};
```

The Java backend (`src/commands/java/dataize.js:26`) passes `...args` through to the underlying `dataize` call unfiltered, with no such filter.

Verified: running `eoc -l js --alone dataize foo.bar -- -x` reaches the eo2js backend with `-x` silently removed from the argument list, and no warning is printed. The identical invocation with the Java backend forwards `-x` unchanged. An EO program that reads its own arguments (e.g. via `args.at 0`) sees a different argument list depending solely on which `--language` backend is selected, with no indication anything was dropped.

## What should happen

The `js` backend should pass `args` through unchanged, the same as the Java backend does, so program arguments are treated consistently across both platforms. If there is a real reason `-`-prefixed arguments cannot reach the `js` backend (e.g. they conflict with an internal flag), that limitation should at least produce a visible warning instead of a silent drop.

## Steps to reproduce

1. Run `eoc -l js --alone dataize <obj> -- -x` on a project whose object reads its arguments.
2. Compare with `eoc -l java --alone dataize <obj> -- -x`.
3. Observe the `-x` argument reaches the Java backend but not the `js` backend.

morphqdd:
The `js` backend dropped every argument starting with a dash before handing them on, and printed nothing about it. The Java backend passes `...args` through unchanged, so an EO program that reads its own arguments saw a different list depending only on which `--language` was picked.

Both backends now pass them through the same way. The filter was not guarding anything: the arguments go to `execFileSync` as argv entries, so a dash in one of them is not read as a flag of ours.

Closes #1123

### DIFF
--- a/src/commands/js/dataize.js
+++ b/src/commands/js/dataize.js
@@ -14,7 +14,7 @@ const eo2jsw = require('../../eo2jsw');
  */
 module.exports = function(obj, args, opts) {
   return eo2jsw(
-    ['dataize', obj, ...args.filter((arg) => !arg.startsWith('-'))],
+    ['dataize', obj, ...args],
     {...opts, alone: true, project: 'project'}
   );
 };


yegor256:

---

PR #1229, opened 2026-08-31
#1168: keep the pinned version when Maven Central cannot be reached

### LINKED ISSUE #1168: `eoc --latest` crashes with a raw stack trace when Maven Central is unreachable
## What happens

`src/parser-version.js`'s `get()` (lines 16-29) does a synchronous request to Maven Central and neither catches transport errors nor handles a non-200 response gracefully - it just throws. It is called from `src/eoc.js:115` at module load time, which is before `program.parseAsync()` runs and therefore outside the top-level `try/catch` at `src/eoc.js:439-446`. So nothing in the process can catch whatever it throws.

Its sibling function `exists()` was explicitly hardened for exactly this scenario in #998 and #1085 (warn and proceed instead of crashing), but `get()` never received the same treatment.

Reproduced in a network namespace with only external network removed (loopback still up):

```
$ unshare -r -n sh -c 'ip link set lo up; node src/eoc.js --latest lint'
node_modules/sync-rpc/lib/index.js:167
    throw error;
    ^
Error
    at extractValue (.../sync-rpc/lib/index.js:165:19)
    at Object.get (src/parser-version.js:20:17)
    at Object.<anonymous> (src/eoc.js:115:40)
    ...
  code: 'ENETUNREACH'
```

Note the thrown error has an empty `message`, so even moving the call inside a try/catch would only print a blank line - the fix needs to be inside `get()` itself, catching the transport error and falling back the same way `exists()` does.

## What should happen

`eoc --latest` should behave gracefully when Maven Central is unreachable - warn and fall back to the bundled parser version (mirroring what `exists()` already does), not print a raw Node internal stack trace with an empty error message.

morphqdd:
`get()` threw on a transport error and on any non-200. It is called at module load, before `parseAsync` runs, so the `try/catch` around the command line never sees it and the user gets a raw stack trace from inside `sync-rpc`.

It now warns and answers with nothing, the way `exists()` next to it was hardened in #998 and #1085, and `--latest` keeps the version the repository pins.

Closes #1168

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -113,7 +113,7 @@ const path = require('path'),
   jeo = fs.readFileSync(path.join(__dirname, '../jeo-version.txt'), 'utf8').trim();
 let parser = fs.readFileSync(path.join(__dirname, '../eo-version.txt'), 'utf8').trim();
 if (process.argv.includes('--latest')) {
-  parser = require('./parser-version').get();
+  parser = require('./parser-version').get() || parser;
   // Maybe here we should also go to GITHUB, find out what is the
   // latest hash of the objectionary/home repository, and then
   // set it to the "hash" variable?
--- a/src/parser-version.js
+++ b/src/parser-version.js
@@ -8,22 +8,36 @@ const colors = require('colors');
 const request = require('sync-request'),
 
   /**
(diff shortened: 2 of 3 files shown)


yegor256:

---

PR #1230, opened 2026-08-31
#1169: say that the JavaScript platform has no such option

### LINKED ISSUE #1169: `--object`, `--stack` and `--heap` are silently discarded on the JavaScript platform
## What happens

`src/eoc.js:321-335` declares `--stack`, `--heap` and `--object` on the `test` command with no platform qualification, and `java/test.js` honours all three on the Java platform. On the JavaScript platform, `src/commands/js/test.js:14` and `src/commands/js/dataize.js:15-19` pass the same options through to `src/eo2jsw.js`, whose `flags()` (lines 15-24) builds its Maven-style flag list from a fixed allow-list (`--target`, `--project`, `--foreign`, `--resources`, `--alone`, `--tests`) and silently drops everything not on that list - no warning is printed.

Verified by stubbing `child_process.execSync` to print the composed command:

```
input: {object: 'foo.app.works', stack: '99M', heap: '9G'} for `test`
EO2JS CMD: node .../eo2js.js test --target .eoc --project project --foreign eo-foreign.json \
  --resources .../resources --alone
   ^ object/stack/heap never reached eo2js

input: ('foo', ['--flag','bar'], {stack:'99M', heap:'9G'}) for `dataize`
EO2JS CMD: node .../eo2js.js dataize foo bar --target .eoc --project project ...
   ^ same for dataize
```

So `eoc -l js test --object foo.app.works` silently runs the entire test suite instead of filtering to one test, with no message telling the user their option was ignored. `eo2js`'s own `test` command genuinely has no per-object filter (only `--exclude`), so there is no way to honour `--object` on the JS platform - but the CLI should say so rather than pretend the option was accepted.

This is the same underlying class as the already-accepted #1123 (which covered only `-`-prefixed arguments being dropped in `js/dataize.js`), just for a different set of options.

## What should happen

When `--object`, `--stack`, or `--heap` are given on the JavaScript platform, `eoc` should print a clear "not supported on the JS platform" message (or error out) instead of silently accepting and then discarding them.

morphqdd:
`--object`, `--stack` and `--heap` are declared on commands that both platforms serve, and `eo2jsw` builds its flags from a fixed list, so on the JavaScript platform all three were dropped without a word. `eoc -l js test --object foo.app.works` ran the whole suite as if nothing had been asked.

`eo2js` has no per-object filter and runs no JVM, so there is nothing to honour. What was missing was saying so. The check reads `process.argv` rather than the parsed options, because `--stack` and `--heap` carry defaults and would otherwise warn on every run; the same idiom is used for `--latest` in `eoc.js`.

Closes #1169

### DIFF
--- a/src/eo2jsw.js
+++ b/src/eo2jsw.js
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: MIT
  */
 
+const colors = require('colors');
 const path = require('path');
 const {execFileSync} = require('child_process'),
 
@@ -31,6 +32,14 @@ const {execFileSync} = require('child_process'),
    * @return {Promise<Array.<String>>}
    */
   eo2jsw = function(command, args) {
+    const dropped = ['object', 'stack', 'heap'].filter(
+      (name) => process.argv.includes(`--${name}`)
+    );
+    if (dropped.length > 0) {
+      console.warn(colors.yellow(
+        `The JavaScript platform has no ${dropped.map((name) => `--${name}`).join(' and no ')}, ignoring`
+      ));
+    }
     const lib = path.resolve(__dirname, '../node_modules/eo2js/src'),
       bin = path.resolve(lib, 'eo2js.js'),
       cmd = Array.isArray(command) ? command : [command];


yegor256:

---

PR #1232, opened 2026-08-31
#1167: let commander parse the object and the arguments

### LINKED ISSUE #1167: `dataize` derives object name and arguments from raw `program.args`, breaking on `--heap`/`--stack` ordering
## What happens

The `dataize` command derives its object name and program arguments directly from raw `program.args` (`src/eoc.js:311-317`: `coms().dataize(program.args[1], program.args.slice(2), ...)`). The command declares `--stack` and `--heap` options (`src/eoc.js:304-305`) but no `.argument()`, so commander leaves those option tokens sitting inside `program.args` instead of parsing them out, and the code indexes into that array blindly.

Verified with a fake `java` on PATH that echoes its argv:

```
$ node src/eoc.js --alone dataize --heap 1G foo
FAKE-JAVA-ARGS: ... -Xmx1G -jar .../eoc.jar --heap 1G foo
   ^ the object name became "--heap" instead of "foo"

$ node src/eoc.js --alone dataize foo --heap 1G
FAKE-JAVA-ARGS: ... -Xmx1G -jar .../eoc.jar foo --heap 1G
   ^ heap is correctly applied to the JVM, but "--heap 1G" also leaks into the
     dataized program's own argv

$ node src/eoc.js --alone dataize ; echo rc=$?
FAKE-JAVA-ARGS: ... -jar .../eoc.jar
rc=0
   ^ no object name given at all: runs the jar with nothing and exits 0
```

So the natural CLI ordering `eoc dataize --heap 1G myobject` silently picks the wrong object name, the "correct" ordering (options first is what most CLI tools also get wrong the other way, but here even putting the object first still pollutes the target program's own arguments) leaks the option tokens into the dataized program's argv, and a missing object name is accepted without any error.

The same raw-`program.args` idiom also appears in `src/commands/js/link.js:19` (`program.args[0] === 'test'`) - it happens to be safe there today given the current option set, but it is the same fragile pattern and could break the same way if a new option is ever added to that command.

## What should happen

`dataize` should declare its operands properly (e.g. `.argument('<object>')` and `.argument('[args...]')`) and use the values commander hands to the action callback, instead of reading `program.args` directly. That would make `--heap`/`--stack` position-independent, keep them out of the dataized program's own argv, and let commander itself report a clear error when the object name is missing.

morphqdd:
The command declared `--stack` and `--heap` but no arguments, so commander left those tokens sitting in `program.args` and the code indexed into that array blindly. `eoc dataize --heap 1G foo` took `--heap` for the object name; with the flag written after the name it reached the JVM but also leaked into the program's own argv; and `eoc dataize` with nothing at all ran the jar with no object and exited 0.

Declaring `<object>` and `[args...]` hands all three cases to commander: the options are parsed out wherever they stand, what is left is the program's own arguments, and a missing object name is refused by name.

The duplicated call in the two branches is gone too, since only the pipeline before it differed.

Closes #1167

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -302,21 +302,17 @@ program.command('link')
 
 program.command('dataize')
   .description('Run the single executable binary and dataize an object')
+  .argument('<object>', 'Full EO name of the object to dataize')
+  .argument('[args...]', 'Arguments to give to the program being dataized')
   .option('--stack <size>', 'Set stack size for the virtual machine', '64M')
   .option('--heap <size>', 'Set the heap size for the VM', '256M')
-  .action(async (str, opts) => {
+  .action(async (object, args, str) => {
     pin(program.opts());
     clear(str);
     if (program.opts().alone === undefined) {
       await pipe()(coms(), ['register', 'assemble', 'lint', 'resolve', 'transpile', 'compile', 'link'], program.opts());
-      await coms().dataize(
-        program.args[1], program.args.slice(2), {...program.opts(), ...str}
-      );
-    } else {
-      await coms().dataize(
-        program.args[1], program.args.slice(2), {...program.opts(), ...str}
-      );
     }
+    await coms().dataize(object, args, {...program.opts(), ...str});
   });
 
 program.command('inspect')


yegor256:

---

PR #1235, opened 2026-09-01
#1233: honour the home tag and merge package members before transpiling

### LINKED ISSUE #1233: `build` and `itest` jobs failing due to `home-tag.txt` being ignored and `eo:merge` not executing
## What happens

Every `build` and `itest` job is red, on master and on every pull request,
including ones that touch nothing but a dependency version (#1211 and #1207,
both `renovate` bumps, fail the same six jobs).

```
[ERROR] Failed to parse '.../temp/test-compile/simple/target/2-pull/bytes.eo:25':
        [25:8] error: 'name must start with a lowercase letter'
```

There are two separate causes, and both are ours.

## 1. `home-tag.txt` is ignored

The tag reaches the plugin as `-Deo.tag=0.62.1`, and `MjSafe` of the pinned
plugin `0.62.1` builds the hash in a field initializer:

```java
protected CommitHash hash = new ChCached(
    new ChNarrow(new ChRemote(this.tag))
);
```

`new ChRemote(this.tag)` runs during construction, before Maven injects
`eo.tag` by reflection, so it captures the `"master"` default and keeps it.
There is no `eo.hash` property to set instead. Objectionary/eo fixed this in
#7253 (`862f4f9893`, 21-Aug), and `git tag --contains` puts that commit in
`0.63.0` only.

So the build has always pulled `home@master`. That was harmless while master
carried the same objects as the tag. On 27-Aug `0.63.0` was released, home's
master moved to `4ddf90b`, and `objects/bytes.eo` gained a line the pinned
parser cannot read:

```
  [] > and /Q.bytes
    ? > ^          <-- line 25, column 8
    ? > b /Q.bytes
```

`home@0.62.1` (`fa95dd7`) has no such line. That explains the timing exactly:
the last green `itest` on master was 25-Aug, the next one was today and it
failed, with nothing in between.

The pulled objects land in `~/.eo/pulled/0.62.1/4ddf90b`, which is master's
hash, while the same run under plugin `0.63.0` puts them in `.../fa95dd7`,
the tag's. That is the whole proof.

## 2. `eo:merge` is never run

Bumping the plugin restores the tag and every object parses, and then the
generated Java does not compile:

```
[ERROR] .../generated-sources/org/eolang/EO_string/EOregex.java:[1296,24] cannot find symbol
[ERROR]   symbol: class EOregex$EOcompile
```

`eo-runtime-0.63.0.jar` ships that atom as
`org/eolang/EOstring$EOregex$EOcompile.class`, nested under `EOstring`, while
`string/regex.eo` is a member of the `string` package and transpiles to an
`EO_string` package of its own.

`MjMerge` is what reconciles the two: "put every member of a package inside
the object that the package names", running after `eo:lint` and before
`eo:transpile`. `eo-runtime` builds itself with `register, format, compile,
inference, merge, transpile`, and our pipeline is `register, assemble, lint,
resolve, place, transpile`. The goal appears nowhere in this repository, so
package members stay split and name classes the runtime does not carry.

It could not have been called before: `eo:merge` arrives in `74e2d32520`
(`bug(#6656)`), and `git tag --contains` gives `0.63.0` alone.

## What I measured

Same one-object program, linting off, `~/.eo` cleared between runs:

| plugin | `home-tag` | `eo:merge` | pulled | parse errors | javac errors | exit |
|---|---|---|---|---|---|---|
| 0.62.1 | 0.62.1 | no | `4ddf90b` (master) | 8079 | - | 1 |
| 0.63.0 | 0.62.1 | no | `fa95dd7` (the tag) | 2 | 0 | 1 |
| 0.63.0 | 0.63.0 | no | `4ddf90b` (the tag) | 0 | 4 | 1 |
| 0.63.0 | 0.63.0 | yes | `4ddf90b` (the tag) | 0 | 0 | **0** |

The last row is a real build, not a skipped one: 1859 `.class` files, a
`4-merge` directory holding `bytes.xmir`, `number.xmir`, `string.xmir` and
`tuple.xmir`, and
`classes/org/eolang/EOstring$EOregex$EOpattern$EOmatch$EOmatched_from_index.class`
among the output, which is one of the two atoms the third row cannot find.

## What should happen

Two changes, and the second needs the first:

1. Raise `eo-version.txt` and `home-tag.txt` to `0.63.0`, so that
   `home-tag.txt` is honoured rather than silently ignored. Right now that
   file is decoration.
2. Add `eo:merge` to the pipeline, between `eo:lint` and `eo:transpile`, the
   way `eo-runtime` orders it.

I checked this on one minimal program, so it fixes the failure that stops the
build; whether every `itest` scenario goes green with it is not something one
program can show.

morphqdd:
Two changes, the second needs the first.

`home-tag.txt` was ignored: plugin 0.62.1 builds the commit hash in a field
initializer, before Maven injects `eo.tag`, so every build pulled home@master.
Raising both files to 0.63.0 makes the tag honoured again, since the fix for
that landed in 0.63.0.

With the tag back, the generated Java still did not compile: `string/regex.eo`
is a member of the `string` package and transpiles into an `EO_string` package
of its own, while the runtime jar carries the atom nested under `EOstring`.
`eo:merge` is what puts a package member inside the object the package names,
and it was never called here. It now runs right before `eo:transpile`, the way
`eo-runtime` orders it.

Closes #1233


morphqdd:
@yegor256, please review

yegor256:
CI is green and the core fix (honouring `home-tag.txt`, adding `eo:merge` before transpile) looks correct and well-tested. Flagging for a manual look rather than auto-merging because the version bump also changes alias-expansion behavior in `test_format.js` (e.g. a self-named alias now emits `Q.math.plus` instead of being dropped, and single-use aliases are now dropped rather than expanded) — that ripple effect isn't mentioned in the PR description and goes beyond the stated scope, so I'd rather a human confirm it's expected before merging.

---

morphqdd:
@yegor256, review please

morphqdd:
@yegor256, please review this pr

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+7059](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1247, opened 2026-09-03
#1239: pass the foreign flag as two argv entries

### LINKED ISSUE #1239: `eoc -l js` fails due to `--foreign eo-foreign.json` being passed as single argv token
## What happens

`flags()` builds an argv array, and one entry still holds a flag and its value joined by a space:

https://github.com/objectionary/eoc/blob/master/src/eo2jsw.js#L19

```js
'--foreign eo-foreign.json',
```

Commit a3b3f8f (PR #1150, fixing #1145) switched `eo2jsw` from `execSync` with a shell string to
`execFileSync(process.execPath, [bin, ...cmd, ...flags])`. Under a shell the space split that
token in two, so it worked. With `execFileSync` there is no shell, so `eo2js` gets one literal
argument `--foreign eo-foreign.json` and its commander rejects it. Every JS-platform command
fails immediately:

```
$ cd /tmp/t1 && node /tmp/zc-eoc/src/eoc.js -l js --alone transpile; echo "EXIT=$?"
error: unknown option '--foreign eo-foreign.json'
Command failed: .../node .../eo2js.js transpile --target .eoc --project project --foreign eo-foreign.json --resources .../resources --alone
EXIT=1
```

The message is misleading on top of that, because the failure text re-joins argv with spaces, so
the printed command looks correct. With the flag split into two entries it gets past option
parsing and reaches real work:

```
$ node .../eo2js.js transpile --target .eoc --project project --foreign eo-foreign.json --resources .../resources --alone
Reading foreign tojos from: /tmp/t1/.eoc/eo-foreign.json
File /tmp/t1/.eoc/eo-foreign.json is not found
```

## What should happen

`'--foreign', 'eo-foreign.json'` as two array elements, the way every other flag in the same list
is written.

morphqdd:
The flag and its value were joined by a space in one array entry. Under the old
shell invocation the space split them; `execFileSync` has no shell, so `eo2js`
got one literal argument and rejected it, which broke every command on the
JavaScript platform.

Closes #1239

### DIFF
--- a/src/eo2jsw.js
+++ b/src/eo2jsw.js
@@ -16,7 +16,7 @@ const {execFileSync} = require('child_process'),
     return [
       '--target', args.target,
       '--project', args.project || 'project',
-      '--foreign eo-foreign.json',
+      '--foreign', 'eo-foreign.json',
       '--resources', path.resolve(lib, 'resources'),
       args.alone ? '--alone' : '',
       args.tests ? '--tests' : ''
@@ -49,3 +49,4 @@ const {execFileSync} = require('child_process'),
   };
 
 module.exports = eo2jsw;
+module.exports.flags = flags;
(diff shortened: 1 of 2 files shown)


yegor256:

---

PR #1248, opened 2026-09-03
#1243: honour the alone option in latex and normalize

### LINKED ISSUE #1243: `--alone` option not being respected by `latex` and `normalize` commands
## What happens

`--alone` is documented as "Just run a single command without dependencies", and every pipeline
command guards its pipeline with `if (program.opts().alone === undefined)`. Two do not:

https://github.com/objectionary/eoc/blob/master/src/eoc.js#L428
https://github.com/objectionary/eoc/blob/master/src/eoc.js#L437

`latex` and `normalize` call `pipe()(coms(), ['register', 'parse'], ...)` unconditionally:

```
$ cd /tmp/t1 && for c in latex normalize print parse; do echo "### --alone $c"; node -r /tmp/stub3.js /tmp/zc-eoc/src/eoc.js --alone $c 2>&1 | grep '>>> mvn'; done
### --alone latex
>>> mvn goals: ["eo:register","eo:parse"]
>>> mvn goals: ["eo:latex"]
### --alone normalize
>>> mvn goals: ["eo:register","eo:parse"]
### --alone print
>>> mvn goals: ["eo:print"]
### --alone parse
>>> mvn goals: ["eo:parse"]
```

So `--alone latex` still pays for a full re-register and re-parse, which is the one thing the flag
exists to avoid.

## What should happen

Wrap both pipeline calls in the same `if (program.opts().alone === undefined)` guard the other
commands use.

morphqdd:
`latex` and `normalize` ran `register` and `parse` unconditionally, so `--alone`
still paid for a full re-register and re-parse, which is the one thing the option
exists to avoid. They now use the same guard every other pipeline command uses.

The test replaces Maven with a stub that prints the goals, so it needs neither
Maven nor the network.

Closes #1243


yegor256:

---

PR #1249, opened 2026-09-03
#1242: clear the directory in the hook, for every command

### LINKED ISSUE #1242: '--clean' option doesn't delete .eoc directory with 'audit', 'foreign', 'register', 'docs', 'generate_comments', 'jeo:disassemble', 'jeo:assemble' commands
## What happens

`-c, --clean` is a program-wide option described as "Delete .eoc directory before running a
command" (`src/eoc.js:152`), but `clear()` is called from inside individual actions rather than
from the `preAction` hook:

https://github.com/objectionary/eoc/blob/master/src/eoc.js#L479-L483

Fourteen actions call it. These seven do not: `audit` (173), `foreign` (181), `register` (197),
`docs` (357), `generate_comments` (383), `jeo:disassemble` (401), `jeo:assemble` (419).

```
$ cd /tmp/t1 && rm -rf .eoc && mkdir .eoc && echo junk > .eoc/leftover.txt
$ node -r /tmp/stub3.js /tmp/zc-eoc/src/eoc.js -c register >/dev/null 2>&1; ls .eoc
leftover.txt
$ rm -rf .eoc && mkdir .eoc && echo junk > .eoc/leftover.txt
$ node -r /tmp/stub3.js /tmp/zc-eoc/src/eoc.js -c jeo:disassemble >/dev/null 2>&1; ls .eoc
leftover.txt
$ rm -rf .eoc && mkdir .eoc && echo junk > .eoc/leftover.txt
$ node -r /tmp/stub3.js /tmp/zc-eoc/src/eoc.js -c parse >/dev/null 2>&1; ls .eoc
ls: cannot access '.eoc': No such file or directory
```

`/tmp/stub3.js` is a preload replacing the Maven spawn with a stub that exits zero, so no Java is
needed. `docs` uses no Maven and behaves the same with no stub at all.

Whether a documented global option does anything depends on which command you typed, with no
warning either way.

## What should happen

Move `clear()` into the `preAction` hook next to the `--dir` handling at `158-169`, so a
program-wide option behaves the same for every command.

morphqdd:
`--clean` is a program-wide option but `clear()` was called from inside fourteen
of the twenty-one actions, so whether it did anything depended on which command
was typed: `audit`, `foreign`, `register`, `docs`, `generate_comments` and both
`jeo` commands ignored it silently.

It moved into the `preAction` hook, next to the `--dir` handling, so it behaves
the same for every command.

Closes #1242

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -155,7 +155,7 @@ program
   .option('--pin <version>', 'Fail if eoc version doesn\'t match exactly', version.what)
   .option('--update-snapshots', 'Update snapshots in the local repository if they are outdated');
 
-program.hook('preAction', (command) => {
+program.hook('preAction', (command, action) => {
   const dir = command.opts().dir;
   if (path.resolve(dir) !== process.cwd()) {
     try {
@@ -165,6 +165,9 @@ program.hook('preAction', (command) => {
       process.exit(1);
     }
     console.debug(`Working directory changed to ${process.cwd()}`);
+  }
+  if (command.opts().clean) {
+    coms().clean({...command.opts(), ...action.opts()});
   }
 });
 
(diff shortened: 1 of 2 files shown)


yegor256:

---

PR #1250, opened 2026-09-03
#1244: tick only on an interactive terminal

### LINKED ISSUE #1244: Maven progress ticker inappropriately writes ANSI escape codes into redirected `stdout`
## What happens

The progress ticker writes to `stdout` with no check that `stdout` is a terminal:

https://github.com/objectionary/eoc/blob/master/src/mvnw.js#L221-L225

`print()` does `process.stdout.write(colors.yellow(...))` and then `readline.clearLine` and
`readline.cursorTo`, once a second. Redirecting or piping the output puts cursor-control escapes
into the file or the downstream process:

```
$ cd /tmp/t1 && node /tmp/zc-eoc/src/eoc.js print > out.txt 2> err.txt; echo "EXIT=$?"
EXIT=1
$ cat -A out.txt
^[[33m[eo:print] 0ms; 0 files generated so far...^[[39m^[[0K^[[1G^[[2K
```

The only thing in the redirected stdout is escapes and a progress line. `--batch` suppresses it,
but nothing does so automatically for a non-interactive stream, which is the normal case in CI.

## What should happen

Gate `start()` on `process.stdout.isTTY` as well as on `batch`, the way the `colors` package
already gates colouring, so a redirected or piped stdout gets no cursor control at all.

morphqdd:
The progress ticker writes cursor-control escapes once a second with no check
that stdout is a terminal, so a redirected or piped run, which is the normal case
in CI, got the escapes in the file or in the downstream process.

The ticker is gated on `process.stdout.isTTY` now, and the two identical `close`
handlers that differed only in whether they stopped it became one.

Closes #1244

### DIFF
--- a/src/mvnw.js
+++ b/src/mvnw.js
@@ -23,6 +23,19 @@ module.exports.summary = function(args) {
 };
 
 /**
+ * Whether the progress ticker should run: it writes cursor-control
+ * escapes, so it only makes sense on an interactive terminal.
+ * @param {Array} args The arguments passed to Maven
+ * @param {String} tgt The target directory, or undefined
+ * @param {Boolean} batch Whether the run is non-interactive
+ * @return {Boolean} TRUE if the ticker should run
+ */
+module.exports.ticking = function(args, tgt, batch) {
+  return tgt !== undefined && args.includes('--quiet') &&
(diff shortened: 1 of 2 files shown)


yegor256:

---

PR #1251, opened 2026-09-03
#1245: keep the Maven failure and drop the stale hint

### LINKED ISSUE #1245: `eoc lint` masks Maven failure details and persistently advises `--easy` even when already enabled
## What happens

Both catch blocks throw a fresh error and hang the original off `cause`:

https://github.com/objectionary/eoc/blob/master/src/commands/lint.js#L28-L33

The top-level handler at `src/eoc.js:460` prints only `err.message`, so the Maven command line and
exit code that every other command shows are lost. And the advice is unconditional: with `--easy`
already on, `extras()` at `lint.js:71` has set `-Deo.failOnWarning=false`, so the failure cannot be
a warning, yet the message still recommends `--easy`.

With a stub that makes the Maven spawn exit 1:

```
$ cd /tmp/t1 && node -r /tmp/stub5.js /tmp/zc-eoc/src/eoc.js --alone --easy lint; echo "EXIT=$?"
There are errors and/or warnings; you may disable warnings via the --easy option
EXIT=1

$ node -r /tmp/stub5.js /tmp/zc-eoc/src/eoc.js --alone parse; echo "EXIT=$?"
The command "/tmp/zc-eoc/mvnw/mvnw eo:parse -Deo.version=0.62.1 ... exited with #1 code
EXIT=1
```

## What should happen

Print the cause along with the message, or let the original error through and append the hint, and
drop the `--easy` suggestion when `opts.easy` is already set. The two catch blocks differ only in
their success message and could be one.

morphqdd:
Both catch blocks threw a fresh error with the original only on `cause`, and the
top-level handler prints just the message, so the Maven command line and exit code
that every other command shows were lost. The `--easy` advice was unconditional
too, so it was printed even when `--easy` was already on and the failure therefore
could not be a warning.

The message now carries the original text, and the hint is added only when the
option is not already set. The two blocks differed only in one word, so they are
one now.

Closes #1245


yegor256:
CI is red on the head commit — `build (ubuntu-24.04, 22, 18)` and `itest (windows-2022, 19, 18, Java)` both failed. The same kind of failure shows up across several other open PRs right now, so it may be a pre-existing/flaky CI issue rather than something this change introduced, but I'd rather have a human confirm than merge on red checks. Reassigning for a look.

---

PR #1252, opened 2026-09-03
#1240: reject when the Maven binary cannot be started

### LINKED ISSUE #1240: `mvnw`'s lack of child `error` event listener leads to uncaught exception and hung promise when Maven is missing
## What happens

`spawn` is given a `close` listener and no `error` listener:

https://github.com/objectionary/eoc/blob/master/src/mvnw.js#L113-L144

When `mvnw/mvnw` is missing, lines 99-102 fall back to `bin = 'mvn'`. If that is not on `PATH`,
`spawn` emits `error`, not `close`. With nothing listening, Node rethrows it as an uncaught
exception, which the `try/catch` around `program.parseAsync` at `src/eoc.js:456-463` cannot catch
because it is an event and not a rejection. The carefully written fallback warning is followed by
a stack dump, and the promise `mvnw()` returned never settles:

```
$ mv /tmp/zc-eoc/mvnw/mvnw /tmp/zc-eoc/mvnw/mvnw.bak
$ cd /tmp/t1 && PATH=/nonexistent $(command -v node) /tmp/zc-eoc/src/eoc.js audit
Warning: mvnw not found at /tmp/zc-eoc/mvnw/mvnw, falling back to system "mvn"
node:events:487
      throw er; // Unhandled 'error' event
      ^
Error: spawn mvn ENOENT
  syscall: 'spawn mvn',
  path: 'mvn',
```

## What should happen

Add `result.on('error', ...)` that rejects with a message naming the binary and saying to install
Maven or repair the package, at the quality of the `javac` diagnostic in `src/jdk.js:31-44`.

morphqdd:
`spawn` was given a `close` listener and no `error` listener. When the bundled
wrapper is missing and no `mvn` is on the PATH, `spawn` emits `error`, so Node
rethrew it as an uncaught exception that the top-level `try/catch` cannot see,
and the promise never settled.

There is an `error` listener now, rejecting with a diagnostic naming the binary
and saying how to install Maven, at the quality of the `javac` message in
`jdk.js`. The two identical `close` handlers became one while there.

Closes #1240


yegor256:

---

morphqdd:
The Windows leg was red because the diagnostic never reached it. `spawn` is given a shell, and a shell answers an unknown command with a non-zero exit rather than with a spawn error, so the `error` event never arrives on Windows and a missing Maven still read as `The command "mvn ..." exited with #1 code`.

`reachable` looks the binary up on the PATH before it is started, honouring `PATHEXT` on Windows, and hands back the same name, so there is no way to get a binary to run without the lookup having happened. The `error` handler stays as a backstop for the cases a lookup cannot see, such as a file that exists but cannot be executed.

Two tests go straight at the lookup, so the behaviour is pinned without depending on how a shell reports a missing command.

PR #1253, opened 2026-09-03
#1241: refuse to write documentation from nothing

### LINKED ISSUE #1241: `eoc docs` generates empty documentation for unparsed projects without error notification
## What happens

`docs` reads XMIR from `path.resolve(opts.target, '1-parse')`:

https://github.com/objectionary/eoc/blob/master/src/commands/docs.js#L127

`findFiles` returns `[]` for a directory that is not there (`src/files.js:16`). Unlike `latex`
(`src/eoc.js:428`) and `normalize` (`437`), the `docs` action never runs the register-and-parse
pipeline first, so on a project that has not been parsed it writes a stylesheet, a `packages.html`
with nothing in it and a summary claiming zero objects, then reports success:

```
$ cd /tmp/t1 && ls
hello.eo
$ node /tmp/zc-eoc/src/eoc.js docs; echo "EXIT=$?"
Summary XML generated at .eoc/docs/summary.xml in 0ms
Documentation generation completed in the /tmp/t1/.eoc/docs directory in 1ms
EXIT=0
$ cat .eoc/docs/summary.xml
<?xml version="1.0" encoding="UTF-8"?>
<eodoc packages="0" objects="0">
</eodoc>
```

Nothing tells the user they had to run `eoc parse` first.

## What should happen

Either run register and parse first, the way `latex` and `normalize` do, or fail with a message
naming the empty directory, instead of exiting zero on documentation nobody can use.

morphqdd:
On a project that had not been parsed, `docs` found no XMIR, wrote a stylesheet,
an empty `packages.html` and a summary claiming zero objects, and exited zero.
Nothing told the user to run `eoc parse` first.

It now fails with a message naming the empty directory, before writing anything.

Closes #1241

### DIFF
--- a/src/commands/docs.js
+++ b/src/commands/docs.js
@@ -9,6 +9,7 @@ const SaxonJS = require('saxon-js');
 const { marked } = require('marked');
 const {elapsed} = require('../elapsed');
 const {findFiles} = require('../files');
+const rel = require('relative');
 
 /**
  * Escape special XML characters.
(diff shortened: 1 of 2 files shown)


yegor256:
CI is red on the head commit — `build (ubuntu-24.04, 22, 18)` and `itest (macos-15, 19, 18, Java)` both failed. The same kind of failure shows up across several other open PRs right now, so it may be a pre-existing/flaky CI issue rather than something this change introduced, but I'd rather have a human confirm than merge on red checks. Reassigning for a look.

---

PR #1254, opened 2026-09-03
#1246: write the mapping the option promises

### LINKED ISSUE #1246: `generate_comments --output` doesn't create a replacement mapping as suggested, but a plain array
## What happens

The option help says the file will contain the replacement mapping
(`src/eoc.js:376-378`), and the write is:

https://github.com/objectionary/eoc/blob/master/src/commands/generate_comments.js#L130

```js
writeFileSync(opts.output, JSON.stringify(results))
```

`generateDocumentation` returns a plain array of model outputs in match order. There is no
mapping: nothing ties a generated comment to the placeholder it belongs to, so a consumer has to
re-run the same regex over the source and trust that the ordering lines up.

```
$ printf 'Describe: {code}\n' > tmpl.txt
$ printf '# <COMMENT-TO-BE-ADDED>\n[] > a\n# <COMMENT-TO-BE-ADDED>\n[] > b\n' > src.eo
$ node /tmp/zc-eoc/src/eoc.js generate_comments --provider placeholder --source src.eo \
    --prompt_template tmpl.txt --output out.json; echo "EXIT=$?"
EXIT=0
$ cat out.json
["<PLACEHOLDER_RESPONSE>","<PLACEHOLDER_RESPONSE>"]
```

## What should happen

Either write a real mapping, carrying each placeholder's index or offset next to its generated
text, which is easy since `generate_comments.js:105-109` already holds `location.index` and throws
it away, or reword the help to say the file holds an ordered list of comments.

morphqdd:
The option help says the file holds the replacement mapping, but a plain array of
model outputs was written, so nothing tied a generated comment to the placeholder
it belongs to and a consumer had to re-run the same regex and trust the ordering.

Each entry now carries the offset of the placeholder it replaces and the
placeholder text next to the comment. The offset was already at hand and thrown
away.

Closes #1246

### DIFF
--- a/src/commands/generate_comments.js
+++ b/src/commands/generate_comments.js
@@ -98,14 +98,19 @@ function getTextFocusedOnSpecificPlaceholder(
  * @param {String} inputCode - Code to replace placeholders in
  * @param {String} commentPlaceholder - Placeholder to replace
  * @param {RunnableSequence} chain - Langchain LLM pipeline
- * @return {Promise<Array.<String>>} of ordered LLM outputs (one for each placeholder in the input)
+ * @return {Promise<Array.<Object>>} of replacements, each carrying the offset of
+ *  the placeholder it replaces and the comment generated for it
  */
 function generateDocumentation(inputCode, commentPlaceholder, chain) {
   const commentPlaceholderRegex = new RegExp(escapeRegExp(commentPlaceholder), 'g');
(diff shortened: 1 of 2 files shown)


yegor256:
CI is red on the head commit — `build (ubuntu-24.04, 22, 18)` and `itest (ubuntu-24.04, 19, 18, Java)` both failed. The same kind of failure shows up across several other open PRs right now, so it may be a pre-existing/flaky CI issue rather than something this change introduced, but I'd rather have a human confirm than merge on red checks. Reassigning for a look.

---

Issue #68, opened 2022-09-19
audit must print parser version too


yegor256:

yegor256:
@rultor release, tag is `0.8.5`

PR #1255, opened 2026-09-03
#1238: report the packages the build depends on

### LINKED ISSUE #1238: `eoc audit` command incorrectly displays Maven version instead of package inspection report
## What happens

`eoc audit` is documented in `README.md` line 119 as:

> `audit` Inspect all packages and report their status

The whole implementation is this:

https://github.com/objectionary/eoc/blob/master/src/commands/audit.js

```js
  return mvnw(['--version'], null, opts.batch);
};
```

It runs `mvnw --version` and prints nothing else, so the output is the Maven banner:

```
Apache Maven 3.9.12 (848fbb4bf2d427b72bdb2471c22fced7ebd9a7a1)
Maven home: /nix/store/pfi6mjf7mi0hl0nzlbipdsycfq0z22p3-maven-3.9.12/maven
Java version: 21.0.12, vendor: N/A, runtime: /nix/store/kvkz322724y8igmcdrnp5cwls6h0mylq-openjdk-headless-21.0.12+2/lib/openjdk
Default locale: en_US, platform encoding: UTF-8
```

No package is inspected and no status is reported. The repository's own test pins exactly that:

```js
it('audits all packages', (done) => {
  const stdout = runSync(['audit']);
  assert(stdout.includes('Apache Maven'), stdout);
  done();
});
```

`git log src/commands/audit.js` shows only licence and lint commits, so the command has never done anything else. #68 asked it to print the parser version as well; that is not printed either, since `mvn --version` reports Maven and Java, not `eo-maven-plugin`.

## Why this is incorrect

A user who reads the README runs `audit` to find out which objects the project pulls, from which `objectionary/home` tag, and whether anything is missing or stale. They get the Maven version instead. The command name and its description promise a check that does not exist, and the test locks the placeholder in place, so the gap cannot be noticed by CI.

## What should happen

Either `audit` reports something about the packages the project actually uses, or the README stops claiming it does.

The information is already reachable. `--home-tag` (default read from `home-tag.txt`, currently `0.62.1`) says which `objectionary/home` revision the build compiles against, `home.objectionary.com/objectionary.lst` lists the objects that revision publishes, and the `+rt` metas of the sources name the JVM artifacts. A useful `audit` would print the resolved home tag, the objects the program refers to, and which of them the registry does not carry, together with the parser version #68 asked for.

morphqdd:
`audit` ran `mvnw --version` and printed nothing else, so it reported the Maven
banner instead of anything about packages, while the README promised a status
report. The test pinned the banner in place.

It now prints the version resolved for this run of every package the build
depends on: eoc itself, the parser, the `objectionary/home` tag and the JEO
plugin, which also covers the parser version asked for in #68. The Maven banner
stays after them. The command description and the README say what it does.

Closes #1238

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ You can also do many other things with `eoc` commands
 (the flow is explained in [this blog post][blog]):
 
 <!-- BEGIN COMMANDS SECTION -->
-* `audit` Inspect all packages and report their status
+* `audit` Print the versions of the packages the build depends on
 * `foreign` Inspect and print the list of foreign objects
 * `clean` Delete all temporary files
 * `register` Register all visible EO source files
--- a/src/commands/audit.js
+++ b/src/commands/audit.js
@@ -3,7 +3,28 @@
  * SPDX-License-Identifier: MIT
  */
 
+const fs = require('fs');
+const path = require('path');
 const {mvnw} = require('../mvnw');
+const version = require('../version');
+
+/**
+ * The packages the build depends on, with the versions resolved for this run.
+ * @param {Hash} opts - All options
+ * @return {Array.<Array.<String>>} of name and version pairs
+ */
+function packages(opts) {
+  return [
+    ['eoc', version.what],
+    ['eo-maven-plugin', opts.parser],
+    ['objectionary/home', opts.homeTag],
+    [
+      'jeo-maven-plugin',
+      opts.jeoVersion ||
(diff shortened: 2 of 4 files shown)


yegor256:
CI is red on the head commit — `build (macos-15, 22, 18)` and `itest (ubuntu-24.04, 19, 18, Java)` both failed. The same two kinds of jobs are failing across several other open PRs right now, so it may be a pre-existing/flaky CI issue rather than something this change introduced, but I'd rather have a human confirm than merge on red checks. Reassigning for a look.

---

PR #1274, opened 2026-09-05
fix(#1259): quote Maven arguments literally for PowerShell

### LINKED ISSUE #1259: On Windows a path containing `$` is silently corrupted before it reaches Maven, because arguments are double-quoted into PowerShell
## What happens

On Windows, `eoc` hands Maven its arguments through PowerShell, inside double quotes, so PowerShell expands what it finds there. A source or target directory whose name contains `$` reaches Maven with a piece missing, and nothing says so.

Reproducing the exact shape of [`src/mvnw.js:113-120`](https://github.com/objectionary/eoc/blob/master/src/mvnw.js#L113-L120) — `spawn(bin, params.map((p) => `"${p}"`), {shell: shell()})` — with `Write-Output` standing in for `mvnw`:

```js
const {spawn} = require('child_process');
const shell = 'C:\\Windows\\SysWOW64\\WindowsPowerShell\\v1.0\\powershell.exe';
const params = [
  '-Deo.sourcesDir=C:\\proj\\a$b\\src',
  '-Deo.targetDir=C:\\proj\\plain\\target'
];
spawn('Write-Output', params.map((x) => `"${x}"`), {stdio: 'inherit', shell: shell});
```

```
-Deo.sourcesDir=C:\proj\a\src
-Deo.targetDir=C:\proj\plain\target
```

The second line is the control and survives. In the first, `$b` is gone: PowerShell read it as a variable, found none, and substituted nothing. Maven is then pointed at a directory that does not exist, and the failure it reports names a path the user never typed.

A backtick behaves the same way — PowerShell's escape character inside double quotes — so `` a`n `` in a path becomes a newline and splits the argument in two.

`$` and `` ` `` are both legal in Windows directory names.

## Root cause

[`src/mvnw.js:29-33`](https://github.com/objectionary/eoc/blob/master/src/mvnw.js#L29) picks PowerShell as the shell on Windows, and line 115 tries to protect the arguments by wrapping each in double quotes:

```js
process.platform === 'win32' ? params.map((p) => `"${p}"`) : params,
```

Double quotes are the wrong quotes for this. In PowerShell they are the *interpolating* form: `$name`, `$(...)` and backtick escapes are all live inside them. Single quotes are the literal form.

Node itself flags the construct. Running the snippet above without `--no-deprecation` prints:

```
[DEP0190] DeprecationWarning: Passing args to a child process with shell option
true can lead to security vulnerabilities, as the arguments are not escaped,
only concatenated.
```

That is the whole mechanism: with `shell` set, `spawn` does not pass an argument vector, it pastes the arguments into one command line and hands it to the shell. Whatever quoting the caller adds has to be correct for that particular shell, and here it is not. Beyond the silent corruption, a directory name containing `$(...)` would be executed rather than read.

## Suggested fix

Single quotes are literal in PowerShell, with a doubled `''` for an embedded quote:

```js
params.map((p) => `'${p.replace(/'/g, "''")}'`)
```

The sturdier fix is to stop going through a shell at all. Nothing in the Maven invocation needs one — there is no pipe, no redirection, no glob — and dropping `shell` lets `spawn` pass the argument vector straight to the process, which removes the quoting question along with the deprecation warning:

```js
const result = spawn(bin, params, {cwd: home, stdio: 'inherit'});
```

`bin` already carries the `.cmd` suffix on Windows at line 98, and Node runs a `.cmd` through the command processor on its own, so the shell option is not what makes `mvnw.cmd` work.

gemshrine:
Fixes #1259.

On Windows the Maven arguments were wrapped in double quotes before being pasted into a PowerShell command line. Double quotes are the interpolating form there, so `$b` in a directory name was read as a variable and dropped, and a backtick was read as an escape.

The arguments are now wrapped in single quotes, the literal form, with an embedded quote doubled. Nothing in a path expands any more.

Tests cover the quoting itself and, on Windows only, a real round trip through `powershell.exe` for a path holding `$`.

### DIFF
--- a/src/mvnw.js
+++ b/src/mvnw.js
@@ -23,6 +23,16 @@ module.exports.summary = function(args) {
 };
 
 /**
+ * Quote a parameter for PowerShell, where single quotes are the literal
+ * form and an embedded quote is doubled.
+ * @param {String} param The parameter to put into the command line
+ * @return {String} The parameter, quoted so that nothing in it expands
+ */
+module.exports.quoted = function(param) {
+  return `'${param.replace(/'/g, `''`)}'`;
+};
+
+/**
  * The shell to use (depending on operating system).
  * @return {String} Path to shell or "undefined" if default one should be used
  */
@@ -112,7 +122,7 @@ module.exports.mvnw = function(args, tgt, batch) {
     console.debug('+ %s', cmd);
     const result = spawn(
       bin,
-      process.platform === 'win32' ? params.map((p) => `"${p}"`) : params,
+      process.platform === 'win32' ? params.map(module.exports.quoted) : params,
       {
         cwd: home,
         stdio: 'inherit',
--- a/test/test_mvnw.js
+++ b/test/test_mvnw.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: MIT
  */
 
-const {mvnw, flags, summary} = require('../src/mvnw');
+const {mvnw, flags, summary, quoted} = require('../src/mvnw');
 const assert = require('assert');
 const fs = require('fs');
 const os = require('os');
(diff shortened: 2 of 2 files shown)


yegor256:

---

PR #1275, opened 2026-09-05
fix(#1260): resolve the Windows shell through SystemRoot

### LINKED ISSUE #1260: The Windows shell path is hard-coded to `C:\Windows\SysWOW64`, ignoring `%SystemRoot%` and naming the 32-bit shell
## What happens

`src/mvnw.js` hard-codes an absolute path to PowerShell:

```js
function shell() {
  if (process.platform === 'win32') {
    return 'C:\Windows\SysWOW64\WindowsPowerShell\v1.0\powershell.exe';
  }
}
```

Two things are wrong with that string, and neither shows up on a default x64 install, which is presumably why it has survived.

It assumes Windows lives on `C:` in a directory named `Windows`. Windows can be installed elsewhere, and the location is what `%SystemRoot%` is for. When it is not `C:\Windows`, `spawn` fails with `ENOENT` naming a path the user never configured, and every `eoc` command that shells out to Maven stops there.

It also names `SysWOW64`, which is the **32-bit** subsystem directory. The native shell on a 64-bit install is under `System32`; both exist on this machine:

```
C:\Windows\SysWOW64\WindowsPowerShell\v1.0\powershell.exe   423424 bytes
C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe   454656 bytes
```

So on x64 the code runs the 32-bit shell to launch a 64-bit toolchain, for no stated reason. On a 32-bit Windows there is no `SysWOW64` at all and the path does not resolve.

## Root cause

The intent is only to name a shell for `spawn`, at line 119. Nothing about the invocation needs a specific bitness or a specific location, so the absolute path is doing more work than the code asked of it.

## Suggested fix

Let Windows resolve it, the way every other tool does:

```js
function shell() {
  if (process.platform === 'win32') {
    return process.env.ComSpec || 'powershell.exe';
  }
}
```

A bare `powershell.exe` is found through `PATH`, which is correct on every install and on both bitnesses. If the location must stay explicit, build it from the environment rather than from a literal:

```js
path.join(process.env.SystemRoot || 'C:\Windows', 'System32', 'WindowsPowerShell', 'v1.0', 'powershell.exe')
```

Worth deciding at the same time whether the shell is needed at all. The Maven call has no pipe, no redirection and no glob, and dropping the `shell` option entirely would remove this question along with the argument-quoting one filed separately.

gemshrine:
Fixes #1260.

The shell was a literal `C:\Windows\SysWOW64\WindowsPowerShell\v1.0\powershell.exe`. That assumes Windows lives on `C:` and names the 32-bit subsystem directory, which does not exist on a 32-bit install at all.

The location now comes from `%SystemRoot%` and points at `System32`, the native shell. When the file is not there the bare `powershell.exe` is left for PATH to resolve, so a machine with an unusual layout still runs.

Tests check the resolved path on Windows and the fallback when the root is wrong, plus that nothing is picked on other platforms.

### DIFF
--- a/src/mvnw.js
+++ b/src/mvnw.js
@@ -23,14 +23,21 @@ module.exports.summary = function(args) {
 };
 
 /**
- * The shell to use (depending on operating system).
+ * The shell to use (depending on operating system). On Windows the location
+ * comes from %SystemRoot%, because Windows is not always installed in
+ * "C:\Windows", and the shell is the native one under System32. When that
+ * file is not there, the bare name is left for PATH to resolve.
  * @return {String} Path to shell or "undefined" if default one should be used
  */
-function shell() {
+module.exports.shell = function() {
   if (process.platform === 'win32') {
-    return 'C:\\Windows\\SysWOW64\\WindowsPowerShell\\v1.0\\powershell.exe';
+    const own = path.join(
+      process.env.SystemRoot || 'C:\\Windows',
+      'System32', 'WindowsPowerShell', 'v1.0', 'powershell.exe'
+    );
+    return fs.existsSync(own) ? own : 'powershell.exe';
   }
-}
+};
 
 let beginning,
   phase = 'unknown',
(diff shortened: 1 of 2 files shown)


yegor256:

---

PR #1276, opened 2026-09-05
fix(#1261): check the metadata before reading a release out of it

### LINKED ISSUE #1261: `parser-version.get()` reads `xml.metadata.versioning.release` unguarded, crashing on an unexpected 200 and yielding `undefined` versions
## What happens

`src/parser-version.js` checks the HTTP status of the Maven Central metadata carefully and then reads the parsed document with no check at all:

```js
if (res.statusCode !== 200) {
  throw new Error(`Invalid response status #${res.statusCode} from ${url}: ${res.body}`);
}
const xml = new XMLParser().parse(res.body);
version.value = xml.metadata.versioning.release;
```

Two different failures come out of those two lines.

When the body is not the document expected — an HTML notice from a proxy or a corporate mirror answering `200`, a truncated response, an empty body — `xml.metadata` is `undefined` and the next property access throws `TypeError: Cannot read properties of undefined (reading 'versioning')`. The user sees a stack trace from inside a version lookup, with no mention of the URL that was fetched or of what came back, while the neighbouring status check exists precisely to produce that kind of message.

When the document parses but carries no `release` element, `version.value` becomes `undefined` and nothing stops it. It travels on into `url(ver)` at line 38 and is interpolated into a request for `.../undefined/eo-maven-plugin-undefined.pom`, and into the `-Deo.version=` flag that `mvnw.js` builds at line 65.

The code already knows this happens. `exists()` guards against it at line 48:

```js
if (ver && ver !== 'undefined') {
```

A comparison against the *string* `'undefined'` is only meaningful for a value that reached this point after being interpolated into text. So the downstream guard was written for a symptom whose source is two functions away and still unguarded.

## Root cause

`get()` treats the parse result as trusted once the status is `200`. A `200` says the server answered, not that it answered with `maven-metadata.xml`.

## Suggested fix

Check what came back before reading through it, and fail with the same shape of message the status check already produces:

```js
const xml = new XMLParser().parse(res.body);
const release = xml?.metadata?.versioning?.release;
if (!release) {
  throw new Error(`No <release> in the metadata at ${url}: ${res.body.toString().slice(0, 200)}`);
}
version.value = String(release);
```

`String(release)` is worth keeping. `fast-xml-parser` converts numeric text by default, so a two-part release such as `1.0` comes back as the number `1` and would be interpolated into a URL that does not exist. Current versions have three parts and are left as strings, which is why this has not bitten yet.

With `get()` unable to yield `undefined`, the `ver !== 'undefined'` comparison in `exists()` becomes dead and can go, leaving the falsy check that actually reads as a guard.

gemshrine:
Fixes #1261.

`get()` checked the HTTP status and then read `xml.metadata.versioning.release` with no check at all. A `200` carrying something other than `maven-metadata.xml` (a proxy notice, a truncated body) came out as `TypeError: Cannot read properties of undefined`, and metadata without a `<release>` quietly produced `undefined`, which then travelled into a POM URL and into `-Deo.version=`.

The document is now checked before it is read through, and the failure names the URL and the beginning of the body, in the shape the neighbouring status check already uses.

`parseTagValue: false` goes with it: `fast-xml-parser` turns a two-part release such as `1.0` into the number `1`, which would be interpolated into a URL that does not exist.

`get()` takes the HTTP call as an argument now, the way `exists()` already did, so the new tests do not need the network.

### DIFF
--- a/src/parser-version.js
+++ b/src/parser-version.js
@@ -9,20 +9,28 @@ const request = require('sync-request'),
 
   /**
    * Load the latest version from GitHub releases.
+   * @param {function} fetch - HTTP call, defaults to sync-request
    * @return {String} Latest version, for example '0.23.1'
    */
   version = module.exports = {
     value: '',
-    get() {
+    get(fetch = request) {
       if (version.value === '') {
         const repo = 'org/eolang/eo-maven-plugin',
           url = `https://repo.maven.apache.org/maven2/${repo}/maven-metadata.xml`,
-          res = request('GET', url, {timeout: 100000, socketTimeout: 100000});
+          res = fetch('GET', url, {timeout: 100000, socketTimeout: 100000});
         if (res.statusCode !== 200) {
           throw new Error(`Invalid response status #${res.statusCode} from ${url}: ${res.body}`);
         }
-        const xml = new XMLParser().parse(res.body);
-        version.value = xml.metadata.versioning.release;
+        const xml = new XMLParser({parseTagValue: false}).parse(res.body),
+          versioning = xml && xml.metadata ? xml.metadata.versioning : undefined,
+          release = versioning ? versioning.release : undefined;
+        if (release === undefined || release === null || release === '') {
+          throw new Error(
+            `No <release> in the metadata at ${url}: ${String(res.body).slice(0, 200)}`
+          );
+        }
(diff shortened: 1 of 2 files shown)


yegor256:

---

PR #1277, opened 2026-09-05
fix(#1262): link the stylesheet relative to the page that carries it

### LINKED ISSUE #1262: `eoc docs` links the stylesheet by absolute filesystem path, so it never loads on Windows and breaks when the docs are moved
## What happens

`eoc docs` writes every generated page with an absolute filesystem path in its stylesheet link, so the styling is lost as soon as the documentation is moved, published, or generated on Windows.

`src/commands/docs.js:130` builds the path:

```js
```

`output` is `path.resolve(opts.target, 'docs')`, so `css` is absolute. It goes straight into the markup at line 90:

```html
```

On Linux that renders as `href="/home/user/project/.eoc/docs/styles.css"`, which resolves only on the machine that generated it. Copy the `docs` directory anywhere else, or serve it over HTTP, and the link points at nothing.

On Windows it is broken from the start. The value is `C:\Users\me\project\.eoc\docs\styles.css`, and a browser reading that as a URL sees no scheme and backslashes where separators should be, so the stylesheet never loads at all, even on the machine that produced it.

The same absolute value is written into pages that live at different depths — line 139 puts a page under `path.dirname(relative)`, one or more directories below `output` — so a relative link has to be computed per page rather than shared.

## Root cause

`css` is used for two different jobs. At line 131 it is a filesystem path, and correctly absolute:

```js
fs.writeFileSync(css, '');
```

At line 90 the same string is a URL, where absolute means something else entirely. Nothing converts between the two.

## Suggested fix

Pass the link relative to the page that carries it, computed from where the page is written:

```js
const link = path.relative(path.dirname(html_app), css).split(path.sep).join('/');
```

`split(path.sep).join('/')` is the part that matters on Windows: a URL wants forward slashes whatever the platform uses. The same value works for the package pages and the summary page, each computed against its own directory.

Since `wrapHtml` and `generatePackageHtml` take `css` as their third argument already, the change is confined to what the callers pass in.

Worth noting that the file is written empty at line 131, so the missing stylesheet changes nothing visible today. That is what has kept the defect quiet, and it will stop being quiet the moment the stylesheet has content.

gemshrine:
Fixes #1262.

Every generated page carried the stylesheet as an absolute filesystem path. On Linux that only resolves on the machine that produced it; on Windows it is `C:\...` with backslashes, which a browser cannot read as a URL at all.

The link is now computed per page, relative to the directory the page is written into, with forward slashes. Object pages sit one or more levels below `docs`, so each gets its own value.

The existing docs test now also asserts the link is relative and holds no drive letter or backslash.

### DIFF
--- a/src/commands/docs.js
+++ b/src/commands/docs.js
@@ -75,10 +75,22 @@ function createXmirHtmlBlock(filepath) {
 }
 
 /**
+ * The stylesheet link of a page, relative to the directory that page
+ * is written into, with forward slashes, since a URL wants them on
+ * every platform.
+ * @param {String} page - Path of the HTML file being written
+ * @param {String} css - Path of the stylesheet
+ * @return {String} The value for the "href" attribute
+ */
+function cssLink(page, css) {
+  return path.relative(path.dirname(page), css).split(path.sep).join('/');
+}
+
+/**
  * Generates Package HTML
  * @param {String} name - Package name
  * @param {String[]} htmls - Array of xmirs htmls
- * @param {String} css - CSS file path
+ * @param {String} css - Stylesheet link, relative to the page
  * @return {String} HTML of the package
  */
 function generatePackageHtml(name, htmls, css) {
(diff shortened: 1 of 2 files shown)


yegor256:

---

PR #1278, opened 2026-09-05
fix(#1263): ask whether a parser version is comparable before comparing it

### LINKED ISSUE #1263: `eoc lint --parser` with a non-semver version dies with a raw `TypeError: Invalid Version`
## What happens

`eoc lint --parser 0.57` dies with a raw `TypeError` from inside `semver`, naming neither the option nor what is wrong with it:

```
TypeError: Invalid Version: 0.57
```

`--parser <version>` takes any string. `src/commands/lint.js:63-68` feeds it straight to `semver.gte`:

```js
function goals(opts) {
  if (opts.parser.endsWith('-SNAPSHOT') || semver.gte(opts.parser, '0.45.0')) {
    return ['eo:lint'];
  }
  return ['eo:verify'];
}
```

`semver.gte` accepts only a strict three-part version and throws on everything else. Checked against `semver` directly:

| `--parser` value | `semver.gte(value, '0.45.0')` |
| --- | --- |
| `0.57.3` | `true` |
| `0.45.0` | `true` |
| `0.57` | **throws** `TypeError: Invalid Version: 0.57` |
| `1.0-rc1` | **throws** `TypeError: Invalid Version: 1.0-rc1` |
| `0.57.3.1` | **throws** `TypeError: Invalid Version: 0.57.3.1` |
| `latest` | **throws** `TypeError: Invalid Version: latest` |
| empty string | **throws** `TypeError: Invalid Version:` |

Only the `-SNAPSHOT` suffix is handled, and only because it is checked first.

## Root cause

The project does validate `--parser`, but later and in another file. `src/mvnw.js:56` asks Maven Central whether the version exists and prints a message naming the option and pointing at the directory listing:

```js
if (opts.parser && !opts.parser.endsWith('-SNAPSHOT') && !parserVersion.exists(opts.parser)) {
```

That check never gets to run for `lint`. `src/commands/lint.js:20` calls `goals(opts)[0]` before line 23 reaches `flags(opts)`, so the raw `TypeError` is thrown first and the friendly diagnostic is unreachable on this path.

## Suggested fix

Ask whether the version is comparable before comparing it:

```js
function goals(opts) {
  const modern = opts.parser.endsWith('-SNAPSHOT')
    || (semver.valid(opts.parser) !== null && semver.gte(opts.parser, '0.45.0'));
  return modern ? ['eo:lint'] : ['eo:verify'];
}
```

`semver.valid` returns `null` instead of throwing, so an unparseable version falls through to `eo:verify` rather than killing the command.

Better still, validate the option once where it enters. `commander` supports a parser argument on `.option()`, and `src/eoc.js` already uses one for `--language` through `canonicalLanguage`. The same treatment for `--parser` would reject a malformed version at parse time, with a message naming the flag, and would let every downstream user of `opts.parser` assume it is well-formed.

gemshrine:
Fixes #1263.

`eoc lint --parser 0.57` died with `TypeError: Invalid Version: 0.57` thrown from inside `semver`, naming neither the option nor what was wrong with it. `semver.gte` accepts only a strict three-part version, and `goals()` handed it whatever the user typed.

`semver.valid` returns `null` instead of throwing, so an unparseable version now falls through to `eo:verify`. The check that already exists in `mvnw.js` then reports the version as unavailable in Maven Central, naming the option and pointing at the directory listing, which is the message the user should have been getting all along.

Tests cover the values that used to throw and the ones that must keep choosing `eo:lint`.

### DIFF
--- a/src/commands/lint.js
+++ b/src/commands/lint.js
@@ -61,7 +61,10 @@ module.exports.goals = goals;
 module.exports.extras = extras;
 
 function goals(opts) {
-  if (opts.parser.endsWith('-SNAPSHOT') || semver.gte(opts.parser, '0.45.0')) {
+  if (
+    opts.parser.endsWith('-SNAPSHOT')
+    || (semver.valid(opts.parser) !== null && semver.gte(opts.parser, '0.45.0'))
+  ) {
     return ['eo:lint'];
   }
   return ['eo:verify'];
--- a/test/commands/test_lint.js
+++ b/test/commands/test_lint.js
@@ -12,6 +12,23 @@ const {runSync, assertFilesExist, parserVersion, homeTag, weAreOnline} = require
 const simple = ['+architect yegor256@gmail.com', '', '[] > simple', ''].join('\n');
 
 describe('lint', () => {
+  it('picks a goal for a version semver cannot parse', () => {
+    for (const parser of ['0.57', '1.0-rc1', '0.57.3.1', 'latest', '']) {
+      assert.deepEqual(
+        lint.goals({parser}),
+        ['eo:verify'],
+        `Expected a goal rather than a TypeError for --parser ${parser}`
+      );
+    }
+  });
+  it('picks the linting goal for a modern parser', () => {
+    assert.deepEqual(lint.goals({parser: '0.57.3'}), ['eo:lint']);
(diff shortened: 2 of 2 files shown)


yegor256:

---

PR #1279, opened 2026-09-05
fix(#1265): state the output limit of phino and name the file that hits it

### LINKED ISSUE #1265: `eoc normalize` fails with non-descriptive buffer error for large XMIR due to inherited 1 MiB `maxBuffer`
## What happens

`eoc normalize` pipes every XMIR through `phino` with `execFileSync` and no `maxBuffer`, so it inherits Node's default cap of 1 MiB on the child's stdout. An XMIR whose normalized form is larger than that kills the command with an opaque error instead of a diagnostic, and the file that broke it is not named.

`src/commands/normalize.js:45-51`:

```js
const out = execFileSync(
  'phino',
  ['rewrite', '--input=xmir', '--output=xmir', '--normalize', xmir],
  {stdio: ['pipe', 'pipe', 'pipe']}
);
```

Node's default for the sync child-process functions is `maxBuffer: 1024 * 1024`. When the child writes more than that, the call throws `ENOBUFS` with `stdout maxBuffer length exceeded`, which says nothing about XMIR, normalization or the file being processed.

The headroom is thinner than it looks. Measuring the parsed XMIR of `eo-runtime` — the standard library, not an unusually large program:

```
571282 bytes   1-parse/directory.xmir
459036 bytes   1-parse/path.xmir
378110 bytes   1-parse/file.xmir
```

The largest is already at 55% of the cap, before normalization, which rewrites the tree and does not generally shrink it. Nothing in the loop bounds what `phino` may emit, so the limit is reached by whichever program crosses it first, and the failure arrives as a buffer error rather than as anything a user can act on.

## Root cause

The cap is Node's default rather than a decision. Nothing in `normalize.js` states a size the pipeline expects, and nothing catches the failure to say which XMIR was being rewritten when it happened. `saveFile(normed, rel, out)` on the next line writes the buffer straight to disk, so the whole document has to fit in memory anyway — the question is only whether the limit is chosen or inherited.

## Suggested fix

State the limit, and name the file when it is hit:

```js
let out;
try {
  out = execFileSync(
    'phino',
    ['rewrite', '--input=xmir', '--output=xmir', '--normalize', xmir],
    {stdio: ['pipe', 'pipe', 'pipe'], maxBuffer: 64 * 1024 * 1024}
  );
} catch (e) {
  throw new Error(`Failed to normalize ${rel}: ${e.message}`, {cause: e});
}
```

64 MiB is two orders of magnitude above the largest XMIR the runtime produces today, which leaves the cap as a guard against a runaway rather than as a ceiling ordinary work runs into.

Writing the child's output straight to a file with `stdio` redirection would remove the cap altogether, since the bytes would never pass through the parent's memory, at the cost of restructuring `saveFile` to take a destination stream.

gemshrine:
Fixes #1265.

The `phino` call inherited Node's default 1 MiB cap on a synchronous child's stdout. The parsed XMIR of `eo-runtime` already reaches 571 KB before normalization, so the cap is close, and crossing it killed the command with `stdout maxBuffer length exceeded` without naming the file being rewritten.

The limit is now stated at 64 MiB, two orders above the largest XMIR the runtime produces, so it guards against a runaway rather than against ordinary work. A failure of the child is wrapped with the XMIR it was rewriting, keeping the original as the cause.

The rewrite moved into a small exported function so the tests can drive it without `phino` installed.

### DIFF
--- a/src/commands/normalize.js
+++ b/src/commands/normalize.js
@@ -18,6 +18,32 @@ const relative = require('relative');
  * @param {Object} opts - All options
  * @return {Promise} of normalize task
  */
+/**
+ * The largest output a single XMIR may produce, in bytes. Node caps the
+ * stdout of a synchronous child at 1 MiB by default, and the parsed XMIR of
+ * the runtime is already halfway there before normalization.
+ */
+const MAX_OUTPUT = 64 * 1024 * 1024;
+
+/**
+ * Rewrite one XMIR through phino.
+ * @param {String} xmir - Path of the XMIR to rewrite
+ * @param {String} rel - Path of that XMIR relative to the parsed directory
+ * @param {function} run - Child process call, defaults to execFileSync
+ * @return {Buffer} The normalized XMIR
+ */
+function normalized(xmir, rel, run = execFileSync) {
+  try {
+    return run(
+      'phino',
+      ['rewrite', '--input=xmir', '--output=xmir', '--normalize', xmir],
+      {stdio: ['pipe', 'pipe', 'pipe'], maxBuffer: MAX_OUTPUT}
+    );
+  } catch (e) {
+    throw new Error(`Failed to normalize ${rel}: ${e.message}`, {cause: e});
+  }
+}
+
 module.exports = function(opts) {
   if (opts.sources === undefined) {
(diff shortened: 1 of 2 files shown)


yegor256:

---

PR #1280, opened 2026-09-05
fix(#1266): refuse a port that is not a TCP port instead of waiting on it

### LINKED ISSUE #1266: `eoc inspect --port` incorrectly retries and fails with an invalid or non-TCP port
## What happens

`eoc inspect --port abc` spends a minute retrying and then blames the server:

```
The inspection server has not opened port NaN in a minute
```

The port never had a chance. `src/commands/java/inspect.js:80` converts it without checking:

```js
const port = Number(opts.port);
```

and `ask()` builds a URL from it on line 40:

```js
answer = await (await fetch(`http://127.0.0.1:${port}/`)).json();
```

A port that is not a valid TCP port makes that URL unparseable, so `fetch` throws before any connection is attempted. `ask` cannot tell that apart from a server that has not opened the port yet, so it sleeps 200 ms and recurses, over and over, until the 60-second deadline on line 91 runs out.

Every one of these fails the same way:

| `--port` | `Number(...)` | `new URL('http://127.0.0.1:<n>/')` |
| --- | --- | --- |
| `8080` | `8080` | ok |
| `abc` | `NaN` | **throws** `TypeError: Invalid URL` |
| `8080abc` | `NaN` | **throws** `TypeError: Invalid URL` |
| `99999` | `99999` | **throws** `TypeError: Invalid URL` |
| `-1` | `-1` | **throws** `TypeError: Invalid URL` |
| empty | `0` | ok — port 0, which means "any free port" |

Meanwhile the JVM was started on line 89 with `--port NaN` and has almost certainly already died, so the minute is spent waiting for a process that is gone.

## Root cause

`--port <number>` is declared at `src/eoc.js:324` with a default of `'8080'` and no parser, so `opts.port` is whatever the user typed. `Number()` turns a bad value into `NaN` or an out-of-range number rather than rejecting it, and the only place the mistake becomes visible is inside a retry loop built for a different failure.

The loop is doing what it should: a connection refused early on is normal, since the JVM needs a moment to bind. What it cannot do is distinguish that from an error that will never resolve.

## Suggested fix

Reject the value where it enters, the way `--language` already is at `src/eoc.js` through `canonicalLanguage`:

```js
.option('--port <number>', 'TCP port for the inspection server', (raw) => {
  const port = Number(raw);
  if (!Number.isInteger(port) || port < 1 || port > 65535) {
    throw new InvalidArgumentError('must be a TCP port between 1 and 65535');
  }
  return port;
}, 8080)
```

`InvalidArgumentError` is already imported in `src/eoc.js`, and commander turns it into a message naming the flag.

Worth separating the two failures inside `ask` as well: a `TypeError` from `fetch` means the request could not be formed and no amount of waiting will change that, while a connection error is the case the retry exists for. Rethrowing the former immediately would keep the minute for the situation that deserves it.

gemshrine:
Fixes #1266.

`eoc inspect --port abc` spent a minute retrying and then reported that the server had not opened port `NaN`. `Number()` turned the value into `NaN`, the request URL could not be parsed, and `ask()` could not tell that apart from a server that has not bound yet.

`--port` now goes through a parser on the option itself, the way `--language` already does, so commander rejects a bad value at parse time with a message naming the flag. `inspect` repeats the check on its own argument, so a caller that skips the CLI is refused before the JVM is started rather than after the deadline runs out.

Tests cover the values from the table in the issue on both paths.

### DIFF
--- a/src/commands/java/inspect.js
+++ b/src/commands/java/inspect.js
@@ -78,6 +78,9 @@ async function ask(port, deadline) {
 module.exports = async function(opts, exec, runner = spawn) {
   verifyJavac(exec);
   const port = Number(opts.port);
+  if (!Number.isInteger(port) || port < 1 || port > 65535) {
+    throw new Error(`Port "${opts.port}" is not a TCP port between 1 and 65535`);
+  }
   const params = [
     '-cp',
     [path.resolve(opts.target, 'eoc.jar'), await jar(opts)].join(path.delimiter),
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -91,6 +91,23 @@ for (const [alias, canonical] of Object.entries(language)) {
  * @return {String} The canonical platform name
  * @throws {InvalidArgumentError} If the language does not resolve to a known platform
  */
+/**
+ * Turn the raw --port value into a TCP port, rejecting anything that is not
+ * one. Without this a value such as "abc" reaches the inspection client as
+ * NaN, makes the request URL unparseable, and is then waited on for a minute
+ * as if the server were slow to start.
+ * @param {String} value - Raw value from the command line
+ * @return {Number} The port
+ * @throws {InvalidArgumentError} If the value is not a TCP port
+ */
+function tcpPort(value) {
+  const port = Number(value);
+  if (!Number.isInteger(port) || port < 1 || port > 65535) {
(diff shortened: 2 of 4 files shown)


yegor256:

---

PR #1281, opened 2026-09-05
fix(#1267): escape an object name on its way into the page

### LINKED ISSUE #1267: `eoc docs` incorrectly interpolates raw package and object names in HTML, resulting in invalid markup
## What happens

`eoc docs` escapes object and package names when it writes them into XML and leaves them raw when it writes the same names into HTML, so a name carrying `&`, `<` or `>` produces malformed pages while the summary beside them stays well-formed.

`src/commands/docs.js:173-175` escapes for the summary:

```js
lines.push(`  <package name="${xmlEscape(pkg)}">`);
for (const obj of info.names) {
  lines.push(`    <object name="${xmlEscape(obj)}"/>`);
}
```

`generatePackageHtml` interpolates the same value with nothing at line 98:

```html
<h1>${name} documentation</h1>
```

and the callers hand it names that come straight from the filesystem. `name` is `path.parse(xmir).name` at line 137, and `package_name` is the directory path with separators replaced by dots at line 144, so both are whatever the source tree is called.

An object in a file named `a&b.eo` yields `<h1>a&b documentation</h1>`, which is not valid HTML — `&b ` is an unterminated entity reference. A browser recovers, so nothing looks broken until the output is fed to anything that parses it strictly.

## Root cause

The helper exists and is applied on one of the two paths. `xmlEscape` was written for this exact hazard and is used four lines apart from the unescaped interpolation, which is what makes the omission easy to miss: the file does not look like it forgot about escaping.

HTML and XML need different escaping in general, but for text going into an element body the two agree on the three characters that matter, so the same helper covers this case.

## Suggested fix

Escape on the way into the markup, in `generatePackageHtml`:

```js
function generatePackageHtml(name, htmls, css) {
  ...
  <h1>${xmlEscape(name)} documentation</h1>
```

`wrapHtml` passes its `name` straight through to `generatePackageHtml`, so the one change covers the per-object pages, the per-package pages and the overall page.

Worth checking `createXmirHtmlBlock` in the same pass. It runs the XMIR through an XSL sheet and then through `marked.parse` at line 58, so whatever the sheet emits is re-parsed as Markdown; a docstring containing Markdown-significant characters is already at the mercy of that step, which is a separate question from the names but lives in the same output.

gemshrine:
Fixes #1267.

Names were escaped on the way into `summary.xml` and left raw on the way into the HTML four lines away, so an object in `a&b.eo` produced `<h1>a&b documentation</h1>`, where `&b ` is an unterminated entity reference.

`generatePackageHtml` now runs the name through the same `xmlEscape` the summary uses. `wrapHtml` passes its name straight through, so the object pages, the package pages and the overall page are all covered by the one change.

### DIFF
--- a/src/commands/docs.js
+++ b/src/commands/docs.js
@@ -95,7 +95,7 @@ function generatePackageHtml(name, htmls, css) {
         <section>
           <header>
             <nav>
-              <h1>${name} documentation</h1>
+              <h1>${xmlEscape(name)} documentation</h1>
               <p>Creation date: ${date.toUTCString()}</p>
             </nav>
           </header>
--- a/test/commands/test_docs.js
+++ b/test/commands/test_docs.js
@@ -48,6 +48,28 @@ describe('docs', () => {
     done();
   });
   /**
+   * Tests that an ampersand in a name reaches the page as an entity.
+   * @param {Mocha.Done} done - Mocha callback signaling asynchronous completion
+   */
+  it('escapes an ampersand in a name', (done) => {
+    fs.writeFileSync(
+      path.join(parsed, 'a&b.xmir'),
+      '<program name="a&amp;b" />'
+    );
+    runSync([
+      'docs',
+      '--verbose',
+      '-s', path.resolve(home, 'src'),
+      '-t', home,
+    ]);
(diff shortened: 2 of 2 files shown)


yegor256:

---

PR #1282, opened 2026-09-05
fix(#1268): hand the object every argument it was given

### LINKED ISSUE #1268: `eoc dataize` inconsistently handles arguments starting with a dash between JavaScript and Java
## What happens

`eoc dataize` passes the object's arguments through on Java and silently throws away any that begin with a dash on JavaScript. The same command line means two different things depending on `--language`.

`src/commands/js/dataize.js:15-20`:

```js
module.exports = function(obj, args, opts) {
  return eo2jsw(
    ['dataize', obj, ...args.filter((arg) => !arg.startsWith('-'))],
  );
};
```

`src/commands/java/dataize.js:20-28` does not filter:

```js
const params = [
  ...
  obj,
  ...args,
].filter((i) => i);
```

So `eoc dataize app -5` hands `-5` to the object under Java and drops it under JavaScript, leaving the object dataized with one argument fewer than the user wrote. Nothing is printed about it. A negative number is the obvious case, and any argument the program itself defines with a leading dash behaves the same way.

`eoc.js:313` splits the command line as `program.args[1]` for the object and `program.args.slice(2)` for its arguments, so everything after the object name is the program's business, not `eoc`'s.

## Root cause

The filter reads as an attempt to keep `eoc`'s own flags out of the object's argument list. Commander has already done that by the time these functions are called — the arguments arrive in `program.args`, past option parsing — so the filter has nothing of `eoc`'s left to remove and only reaches the program's own arguments.

The Java path shows what was meant: pass them along untouched.

## Suggested fix

Drop the filter, so the two languages agree:

```js
module.exports = function(obj, args, opts) {
  return eo2jsw(
    ['dataize', obj, ...args],
  );
};
```

Worth a look at the Java side in the same pass. Its `.filter((i) => i)` exists to drop the empty string that `opts.verbose ? '--verbose' : ''` leaves behind on line 25, but it is applied to the whole array, so an argument that is an empty string is dropped from the program's arguments too. Building the `--verbose` flag conditionally instead of filtering afterwards would keep that from reaching the user's arguments.

gemshrine:
Fixes #1268.

`eoc dataize app -5` handed `-5` to the object on Java and dropped it on JavaScript. The filter reads as an attempt to keep `eoc`'s own flags out, but commander has already taken those out by the time the command is called, so it only reached the program's own arguments.

The filter is gone, and the two languages now agree.

The Java side had a smaller version of the same thing: `.filter((i) => i)` existed to drop the empty string left by `opts.verbose ? '--verbose' : ''`, but it ran over the whole array, so an argument that was an empty string went with it. The flag is built conditionally now and the arguments are concatenated untouched.

The JavaScript command takes its runner as an argument, the way the Java one already did, so both paths are covered by a test.

### DIFF
--- a/src/commands/java/dataize.js
+++ b/src/commands/java/dataize.js
@@ -22,10 +22,7 @@ module.exports = function(obj, args, opts, exec, runner = spawn) {
     `-Xss${opts.stack}`,
     `-Xmx${opts.heap}`,
     '-jar', path.resolve(opts.target, 'eoc.jar'),
-    opts.verbose ? '--verbose' : '',
-    obj,
-    ...args,
-  ].filter((i) => i);
+  ].concat(opts.verbose ? ['--verbose'] : []).concat([obj]).concat(args);
   console.debug(`+ java ${params.join(' ')}`);
   runner('java', params, {stdio: 'inherit'}).on('close', (code) => {
     if (code !== 0) {
--- a/src/commands/js/dataize.js
+++ b/src/commands/js/dataize.js
@@ -10,11 +10,12 @@ const eo2jsw = require('../../eo2jsw');
  * @param {String} obj - Name of object to dataize
  * @param {Array} args - Arguments
  * @param {Object} opts - All options
+ * @param {Function} [run] - Optional runner, defaults to eo2jsw
  * @return {Promise} of executed command
  */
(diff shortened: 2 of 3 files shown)


yegor256:

---

PR #1283, opened 2026-09-05
fix(#1269): give the foreign flag and its value separate argv entries

### LINKED ISSUE #1269: `--foreign eo-foreign.json` flag in eo2js is ineffective due to incorrect argv entry formatting
## What happens

`src/eo2jsw.js` builds the argument vector for `eo2js` and puts one flag and its value into a single entry:

```js
flags = function(args, lib) {
  return [
    '--target', args.target,
    '--project', args.project || 'project',
    '--foreign eo-foreign.json',
    '--resources', path.resolve(lib, 'resources'),
    ...
```

Every other flag is two entries — the name, then the value. `--foreign` is one string with a space in it.

That string is handed to `execFileSync` at lines 39-43:

```js
execFileSync(
  process.execPath,
  [bin, ...cmd, ...flags(args, lib)],
  {timeout: 1200000, windowsHide: true, stdio: 'inherit'}
);
```

`execFileSync` passes the array as `argv` and starts no shell, so nothing splits that entry. `eo2js` receives one argument whose text is `--foreign eo-foreign.json`, which is not the option `--foreign` carrying the value `eo-foreign.json`. The option is never set, so the foreign catalog `eoc` meant to name is not the one `eo2js` uses — it falls back to whatever its own default is.

Every JavaScript-language command goes through this function: `js/transpile`, `js/link`, `js/compile`, `js/dataize`, `js/test` and `js/resolve` all call `eo2jsw`.

The function's own comment says what the shape should be — "Flags for eo2js, **ready to pass as argv entries**" — and the other three flags follow it.

## Root cause

The entry reads as if it were destined for a command line that a shell would split. It is the only one of the four written that way, which suggests it was typed from a shell invocation and never adjusted when the surrounding array was.

Nothing catches it. `filter((flag) => flag !== '')` on line 23 removes empty entries only, and `eo2js` presumably ignores an argument it does not recognise rather than failing, so the flag goes missing quietly.

## Suggested fix

```js
-      '--foreign eo-foreign.json',
+      '--foreign', 'eo-foreign.json',
```

Worth confirming what `eo2js` does today with the unrecognised argument before the change lands: if it has been silently defaulting to the same file all along, the fix is invisible; if it has been reading a different one, the fix changes behaviour and the commands should be re-run against a project that has a foreign catalog.

gemshrine:
Fixes #1269.

`'--foreign eo-foreign.json'` was a single argv entry. `execFileSync` starts no shell, so nothing split it, and `eo2js` received one argument whose text was `--foreign eo-foreign.json` rather than the option carrying the value. The catalog `eoc` meant to name was never the one `eo2js` used.

It is two entries now, like the three flags around it.

`flags` is exported so a test can assert the shape of the vector, including that no entry holds a flag and its value together.

### DIFF
--- a/src/eo2jsw.js
+++ b/src/eo2jsw.js
@@ -16,7 +16,7 @@ const {execFileSync} = require('child_process'),
     return [
       '--target', args.target,
       '--project', args.project || 'project',
-      '--foreign eo-foreign.json',
+      '--foreign', 'eo-foreign.json',
       '--resources', path.resolve(lib, 'resources'),
       args.alone ? '--alone' : '',
       args.tests ? '--tests' : ''
@@ -49,3 +49,5 @@ const {execFileSync} = require('child_process'),
   };
 
 module.exports = eo2jsw;
+
+module.exports.flags = flags;
--- a/test/test_eo2jsw.js
+++ b/test/test_eo2jsw.js
@@ -1,0 +1,24 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+
+const assert = require('assert');
+const {flags} = require('../src/eo2jsw');
+
+describe('eo2jsw', () => {
+  it('puts the name of every flag in its own argv entry', () => {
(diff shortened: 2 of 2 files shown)


yegor256:

---

PR #1284, opened 2026-09-05
fix(#1270): fold the object path into the test class name

### LINKED ISSUE #1270: 'eoc test --object' incorrectly builds package path causing mismatch with transpiled class name
## What happens

`eoc test --object <name>` builds a surefire selector that matches nothing for any object inside a package. Only a top-level object works.

`src/commands/java/test.js:24-34`:

```js
...
const cls = `EO${obj}*Test`;
args.push(`-Dtest=${pkg ? `org.eolang.${pkg}.${cls}` : `org.eolang.${cls}`}#${method}`);
```

The remaining path segments become **package** segments. The transpiler does not put them there. Every generated test class lives directly in `org.eolang`, with the whole path folded into the class name. From a built `eo-runtime`:

```
org/eolang/EOdirectoryEOlistedTest.class
org/eolang/EOchunkEOreadTest.class
org/eolang/EObytesEOsliceTest.class
org/eolang/EOstringEOregexEOpatternEOcheckedTest.class
org/eolang/EOnumberTest.class
```

So for `--object string.regex.compile.some-method` the command asks for

```
-Dtest=org.eolang.EOstring.EOregex.EOcompile*Test#some_method
```

while the class that exists is `org.eolang.EOstringEOregexEOcompileTest`. Surefire matches nothing, and the run reports no tests rather than reporting that the object was not found.

The single-segment case is the one that works, because `parts` is empty after the two `pop()` calls and the `pkg ?` branch falls back to `org.eolang.EOnumber*Test#...`, which does match `EOnumberTest`. That is presumably how the feature was tested.

## Root cause

Two naming conventions are being confused. `EO`-prefixing each segment is right; joining them with `.` is not — the transpiler concatenates them into one identifier and leaves the package alone.

## Suggested fix

Fold the path into the class name instead of into the package:

```js
const cls = parts.map((p) => `EO${p.replace(/-/g, '_')}`).join('');
args.push(`-Dtest=org.eolang.${cls}*Test#${method}`);
```

which gives `org.eolang.EOstringEOregexEOcompile*Test#some_method` for the example above, and leaves the working single-segment case unchanged.

Worth adding a check that the selector matched something. Surefire is content to run zero tests, so a mistyped object name currently looks the same as a passing run; `-DfailIfNoSpecifiedTests=true` turns that into a failure that names what was asked for.

gemshrine:
Fixes #1270.

`eoc test --object string.regex.compile.some-method` asked surefire for `org.eolang.EOstring.EOregex.EOcompile*Test`, treating the leading segments as packages. The transpiler does not put them there: every generated test class lives directly in `org.eolang` with the whole path folded into the identifier, as `EOstringEOregexEOcompileTest`. Nothing matched, and the run reported no tests rather than reporting that nothing was found.

The segments are folded into the class name now. The single-segment case, which happened to work through the `pkg ?` fallback, produces the same selector as before.

Tests drive the command with Maven replaced by a spy and assert the selector for both shapes.

### DIFF
--- a/src/commands/java/test.js
+++ b/src/commands/java/test.js
@@ -28,10 +28,8 @@ module.exports = function(opts, maven = mvnw) {
       );
     }
     const method = parts.pop().replace(/-/g, '_');
-    const obj = parts.pop().replace(/-/g, '_');
-    const pkg = parts.map((p) => `EO${p.replace(/-/g, '_')}`).join('.');
-    const cls = `EO${obj}*Test`;
-    args.push(`-Dtest=${pkg ? `org.eolang.${pkg}.${cls}` : `org.eolang.${cls}`}#${method}`);
+    const cls = parts.map((p) => `EO${p.replace(/-/g, '_')}`).join('');
+    args.push(`-Dtest=org.eolang.${cls}*Test#${method}`);
   }
   return elapsed(async (tracked) => {
     const result = await maven(
--- a/test/commands/test_test.js
+++ b/test/commands/test_test.js
@@ -124,3 +124,36 @@ describe('test', () => {
     done();
   });
 });
+
+describe('test/java', () => {
+  const javaTest = require('../../src/commands/java/test');
+  /**
+   * Run the command with Maven replaced by a spy.
+   * @param {String} object - Value of the --object option
+   * @return {Promise<Array.<String>>} The arguments Maven was called with
+   */
+  async function selector(object) {
+    let seen;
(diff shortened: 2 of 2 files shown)


yegor256:

---

gemshrine:
Superseded by the TestEO* rename in 559cb10 (#1289). The problem of #1270 is still on master, but the right shape changed with it, so I would rather redo this than rebase it. Full analysis in the issue.

PR #1285, opened 2026-09-05
fix(#1271): guard the loop by its key and refuse an unknown level

### LINKED ISSUE #1271: `tinted-console.enable()` improperly validates level due to typing error
## What happens

`enable()` in `src/tinted-console.js` guards its loop with the wrong variable, and the mistake is load-bearing: correcting it the obvious way changes behaviour.

Lines 45-54:

```js
module.exports.enable = function enable(level) {
  for (const key in levels) {
      levels[key] = true;
      if (key === level) {
        break;
      }
    }
  }
};
```

The loop variable is `key`; the guard asks about `level`, the argument. `hasOwnProperty` in this position is meant to skip inherited properties, which is what a linter's `guard-for-in` rule is asking for, and it can only do that if it is given the key being visited.

Because it is given the argument instead, the condition is constant for the whole loop: true when `level` names a real level, false when it does not. The false case is what stops an unknown level from doing anything, so the typo is currently acting as input validation:

```
enable("nope")   with the guard as written  -> {trace:false, debug:false, info:true, warn:true, error:true}
enable("nope")   with the guard corrected   -> {trace:true,  debug:true,  info:true, warn:true, error:true}
```

Correcting `level` to `key`, which is what the code plainly meant, silently turns an unrecognised level into "enable everything", `trace` included. `src/eoc.js:102-107` calls `enable('debug')` from `--verbose` and `--debug`, so today the only inputs are valid; the next caller passing something else would get every level at once and no complaint.

## Root cause

Two separate things are missing and one typo is standing in for both: the loop needs an own-property guard, and the function needs to reject a level it does not know. Written as it is, one wrong identifier happens to cover the second while failing at the first.

## Suggested fix

Separate them, so neither depends on the other:

```js
module.exports.enable = function enable(level) {
  if (!Object.prototype.hasOwnProperty.call(levels, level)) {
    throw new Error(`Unknown logging level: ${level}`);
  }
  for (const key of Object.keys(levels)) {
    levels[key] = true;
    if (key === level) {
      break;
    }
  }
};
```

`Object.keys` returns own enumerable keys, so the guard inside the loop is no longer needed at all, and the check on the way in says out loud what the typo was doing by accident.

Worth noting the neighbouring `before.call(before, ...)` on lines 33-34 while the file is open: the original console method is invoked with itself as `this` rather than with `console`. Node's console methods do not depend on `this`, so nothing breaks today, but `console` is the receiver that was meant.

gemshrine:
Fixes #1271.

`enable()` guarded its loop with `hasOwnProperty(level)` — the argument, not the key being visited — so the condition was constant for the whole loop. It skipped no inherited property, which is what the guard is for, and it happened to reject an unknown level as a side effect.

The two jobs are separated now. An unknown level is refused by name, and the walk goes over `Object.keys`, which returns own enumerable keys, so no guard inside the loop is needed.

Without the check on the way in, correcting the typo would have quietly turned an unrecognised level into "enable everything".

### DIFF
--- a/src/tinted-console.js
+++ b/src/tinted-console.js
@@ -39,16 +39,18 @@ for (const level in levels) {
 }
 
 /**
- * Enable this particular logging level.
+ * Enable this particular logging level, and every level above it.
  * @param {String} level - The level to enable
+ * @throws {Error} If the level is not one of the known ones
  */
 module.exports.enable = function enable(level) {
-  for (const key in levels) {
-    if (levels.hasOwnProperty(level)) {
-      levels[key] = true;
-      if (key === level) {
-        break;
-      }
+  if (!Object.hasOwn(levels, level)) {
+    throw new Error(`Unknown logging level: ${level}`);
+  }
+  for (const key of Object.keys(levels)) {
+    levels[key] = true;
+    if (key === level) {
+      break;
     }
   }
 };
--- a/test/test_tinted_console.js
+++ b/test/test_tinted_console.js
@@ -10,4 +10,31 @@ describe('tinted-console', () => {
     require('../src/tinted-console');
     assert.doesNotThrow(() => console.warn('ok'));
   });
+  it('refuses a level it does not know', () => {
+    const {enable} = require('../src/tinted-console');
+    assert.throws(
+      () => enable('nope'),
+      /Unknown logging level: nope/,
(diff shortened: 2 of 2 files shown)


yegor256:

---

PR #1286, opened 2026-09-05
fix(#1272): leave console.trace alone

### LINKED ISSUE #1272: `tinted-console` incorrectly suppresses `console.trace`, disabling stack trace output
## What happens

`src/tinted-console.js` replaces `console.trace` along with the ordinary logging methods, and turns it off by default. Anything that calls `console.trace` — `eoc`'s own code, or any dependency loaded into the same process — prints nothing at all, and would print no stack trace even when enabled.

Lines 9-15 list it as a level, defaulted to off:

```js
levels = {
  'trace': false,
  'debug': false,
  'info': true,
  ...
```

and lines 25-39 overwrite the global with a suppressible wrapper:

```js
for (const level in levels) {
    const lvl = level, before = console[lvl];
    console[level] = function(...args) {
      if (!levels[lvl]) {
        return;
      }
      before.call(before, safe[colors[lvl]](util.format(...args)));
    };
  }
}
```

`console.trace` is not a severity in Node. It prints the message *and* the stack that led to it, which is its entire purpose. After this module is loaded it does neither: with `trace` false it returns immediately, and once `--verbose` flips it on through `enable('debug')` — which walks the list from the top and switches `trace` on along the way — it prints a gray one-line message, because the wrapper passes an already-formatted string and the original method's stack output is of the call inside the wrapper rather than of the caller.

The module is imported first thing in `src/eoc.js`, before any command loads, so the replacement is in place for the whole process, including third-party code that has no idea `eoc` redefined it.

`trace` is also the only one of the five that `eoc` never calls itself: `console.debug`, `console.info`, `console.warn` and `console.error` all appear across `src/`, and `console.trace` appears nowhere. It was included because it looks like it belongs in a list of levels.

## Suggested fix

Leave it alone:

```js
levels = {
  'debug': false,
  'info': true,
  'warn': true,
  'error': true,
},
```

with the matching entry dropped from `colors`. `enable('debug')` keeps working, since the walk starts at the first key and `debug` is now that key.

If a suppressible trace level is genuinely wanted, it should not take the name of an existing console method that does something else — a `console.verbose` of one's own, or a logger object the code imports, keeps the global intact for everyone else sharing the process.

gemshrine:
Fixes #1272.

`console.trace` was treated as a severity and replaced with a suppressible wrapper, defaulted to off. It is not a severity: it prints the message together with the stack that led to it, and that is the whole point of it. After the module loaded it did neither, for `eoc` and for every dependency sharing the process, since the module is imported first thing in `src/eoc.js`.

`trace` is dropped from the list of levels and from the colours. `enable('debug')` still walks from the first key, which is now `debug`, so `--verbose` behaves as before.

`console.trace` appears nowhere in `src/`, unlike the other four, which suggests it was listed because it looked like it belonged.

### DIFF
--- a/src/tinted-console.js
+++ b/src/tinted-console.js
@@ -7,7 +7,6 @@ const safe = require('colors/safe');
 const util = require('node:util'),
 
   levels = {
-    'trace': false,
     'debug': false,
     'info': true,
     'warn': true,
@@ -15,7 +14,6 @@ const util = require('node:util'),
   },
 
   colors = {
-    'trace': 'gray',
     'debug': 'gray',
     'info': 'white',
     'warn': 'yellow',
--- a/test/test_tinted_console.js
+++ b/test/test_tinted_console.js
@@ -10,4 +10,27 @@ describe('tinted-console', () => {
     require('../src/tinted-console');
     assert.doesNotThrow(() => console.warn('ok'));
   });
+  it('leaves console.trace printing its stack', () => {
+    require('../src/tinted-console');
+    const written = [];
+    const original = process.stderr.write;
+    process.stderr.write = (chunk) => {
+      written.push(String(chunk));
+      return true;
+    };
+    try {
+      console.trace('looking for the caller');
+    } finally {
+      process.stderr.write = original;
+    }
+    const out = written.join('');
+    assert.ok(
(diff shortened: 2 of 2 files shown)


yegor256:

---

PR #1287, opened 2026-09-05
fix(#1273): resolve the latest parser version when a command runs

### LINKED ISSUE #1273: `--latest` option triggers network request even for `--help` and `--version` commands
## What happens

`--latest` fetches from Maven Central while `src/eoc.js` is still being loaded, before commander has looked at the command line. So `eoc --latest --help` and `eoc --latest --version` both block on a network request, and fail with a network error when there is none.

`src/eoc.js:115-119`:

```js
let parser = fs.readFileSync(path.join(__dirname, '../eo-version.txt'), 'utf8').trim();
if (process.argv.includes('--latest')) {
  parser = require('./parser-version').get();
```

This is module scope, not a command action. It runs on import, so it runs for every invocation whose argv contains `--latest`, whatever the rest of the line says — including the two that produce no build at all.

The request is synchronous. `parser-version.get()` uses `sync-request` with `timeout: 100000, socketTimeout: 100000`, so an unreachable or slow repository stalls the process for up to a hundred seconds with nothing printed, since the `console.info` naming the version comes after the call returns.

Offline, it does not stall, it fails. `get()` throws on any non-200, and a failure to connect propagates out of a module-level statement, which means the process dies before commander is reached and before `program.parseAsync` could have turned it into the message on `src/eoc.js:459-461`. Asking `eoc --latest --help` what the flags are is not a request that should need the network at all.

`process.argv.includes('--latest')` is doing option parsing by hand, in a file that has commander configured a few lines below. It also matches the string anywhere on the line, so an object argument that happens to be `--latest` triggers the fetch too.

## Suggested fix

Resolve the version where it is needed rather than where the module loads. The value is only consumed through `program.opts().parser`, so a `preAction` hook — the file already has one on `src/eoc.js:159` for `--dir` — is the natural place:

```js
program.hook('preAction', (command) => {
  if (command.opts().latest) {
    command.setOptionValue('parser', require('./parser-version').get());
  }
});
```

`--help` and `--version` never reach a hook, so both stop touching the network, and a failure to reach Maven Central becomes an error from a command that was actually asked to build something.

Worth reconsidering `sync-request` at the same time. It is deprecated and it blocks the event loop for the whole timeout; the rest of the file is already `async`, and `get()` has exactly one caller.

gemshrine:
Fixes #1273.

The `--latest` fetch sat at module scope, behind `process.argv.includes('--latest')`, so it ran while `src/eoc.js` was still loading — for `eoc --latest --help` and `eoc --latest --version` as much as for a build. The request is synchronous with a 100-second timeout, so a slow repository stalled those two with nothing printed, and an unreachable one killed the process before commander was ever reached.

The version is resolved in a `preAction` hook now, next to the one that handles `--dir`. Help and version never reach a hook, and a failure to reach Maven Central belongs to a command that was actually asked to build something.

This also stops matching `--latest` anywhere on the line, so an object argument of that name no longer triggers a fetch.

Both new tests fail on master and pass here.

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -111,15 +111,7 @@ const fs = require('fs');
 const path = require('path'),
   tag = fs.readFileSync(path.join(__dirname, '../home-tag.txt'), 'utf8').trim(),
   jeo = fs.readFileSync(path.join(__dirname, '../jeo-version.txt'), 'utf8').trim();
-let parser = fs.readFileSync(path.join(__dirname, '../eo-version.txt'), 'utf8').trim();
-if (process.argv.includes('--latest')) {
-  parser = require('./parser-version').get();
-  // Maybe here we should also go to GITHUB, find out what is the
-  // latest hash of the objectionary/home repository, and then
-  // set it to the "hash" variable?
-} else {
-  console.debug(`EO parser ${parser}; use the --latest flag if you need a fresher one`);
-}
+const parser = fs.readFileSync(path.join(__dirname, '../eo-version.txt'), 'utf8').trim();
 
 const version = require('./version');
 program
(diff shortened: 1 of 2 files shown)


yegor256:

---

PR #1296, opened 2026-09-06
#1256: run the case once, it does not vary

### LINKED ISSUE #1256: `test_generate_comments.js` unnecessarily repeats the same test case three times
## What happens

`test/commands/test_generate_comments.js:55-80` wraps its body in a loop whose
variable it never reads:

https://github.com/objectionary/eoc/blob/master/test/commands/test_generate_comments.js#L56

```js
for (let placeholders = 0; placeholders < 3; ++placeholders) {
  const exampleInput = [
    '# <STRUCTURE-BELOW-IS-TO-BE-DOCUMENTED>',
    ...
  ].join('\n');
  ...
  const expectedContents = ['<PLACEHOLDER_RESPONSE>', '<PLACEHOLDER_RESPONSE>'];
  verifyGeneratedOutput(stdout, home, outputFilePath, expectedContents);
}
```

`exampleInput` is a fixed array and `expectedContents` is a fixed pair, so all
three iterations run the same input, spawn `node src/eoc.js` again, and assert the
same thing. The neighbouring test at `:38` has the same loop and does use the
variable, both in the input and in the expectation, which is what this one looks
like it was copied from.

Two of the three subprocesses are pure waste, and it shows in the timing:

```
$ npx mocha test/commands/test_generate_comments.js --timeout 1200000
    ✔ fills output as expected when encountering valid EO code (1679ms)
```

with the loop reduced to one iteration, everything still passing:

```
    ✔ fills output as expected when encountering valid EO code (541ms)
```

It also reads as if the case were parameterised when it is not, so a reader looking
for what varies finds nothing.

## What should happen

Either drop the loop and run the case once, or make the input and the expectation
depend on the loop variable the way the test above it does.

morphqdd:
The loop around this case never reads its variable: the input and the expectation are both
fixed, so all three iterations spawned `node src/eoc.js` with the same arguments and
asserted the same thing. The neighbouring case at `:38` does use the variable, which is
what this one looks like it was copied from.

The case runs once now, 1679ms to 550ms, and it no longer reads as parameterised when it
is not.

Closes #1256

### DIFF
--- a/test/commands/test_generate_comments.js
+++ b/test/commands/test_generate_comments.js
@@ -54,29 +54,27 @@ describe('generate_comments', () => {
   });
   it('fills output as expected when encountering valid EO code', (done) => {
     const home = makeHome();
-    for (let placeholders = 0; placeholders < 3; ++placeholders) {
-      const exampleInput = [
-        '# <STRUCTURE-BELOW-IS-TO-BE-DOCUMENTED>',
-        '[args] > simple',
-        '  io.stdout (args.at 0) > @',
-        '  # <STRUCTURE-BELOW-IS-TO-BE-DOCUMENTED>',
-        '  [] > current-time',
-        '    output. > @',
-        '      sys.posix',
-        '        "gettimeofday"',
-        '        * sys.posix.timeval'
-      ].join('\n');
-      const outputFilePath = path.resolve(home, 'out.json');
-      const stdout = runSync([
-        'generate_comments',
-        '--provider=placeholder',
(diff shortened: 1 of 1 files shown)


yegor256:
Thanks for the contribution, merged!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6792](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1297, opened 2026-09-06
#1290: run itest and grunt on a daily schedule

### LINKED ISSUE #1290: Intermittent `itest` failures due to external dependency changes not detected in real-time
## What happens

A pull request can be green and the same tree red on `master` minutes later, because the
`itest` job depends on a repository nobody here controls and the checks are not re-run
when a pull request is merged.

What happened with #1235, which is what made me notice:

```
#1235 itest, all three legs: success, started 2026-09-01T08:43:36Z
objectionary/home 532094b5 "New release for eo-0.63.1": 2026-09-04T10:46:05Z
#1235 merged:                          2026-09-05T10:52:17Z
master itest:                 failure, 2026-09-05T10:52:20Z
```

`git diff bb851b1a upstream/master` is empty, so the code was the same. The green tick
was four days old when the pull request went in, and `itest` pulls its objects from
`objectionary/home` over the network while it runs, so the answer it gets depends on the
day.

`itest.yml` runs on `push` to `master` and on `pull_request`, and on nothing else. So
between two merges nobody learns that the outside world broke the build; the next
contributor learns it, on a pull request that has nothing to do with it. That is how
#1288 surfaced: three red legs on twenty-two open pull requests at once.

The same holds for `build`, which resolves `eo-maven-plugin` from Maven Central.

## What should happen

Two things, either of which would have caught it on 4 September instead of 5 September,
and both are small:

1. `itest.yml` (and `build`) should also run on a schedule, once a day:

```yaml
'on':
  push:
    branches: [master]
  pull_request:
    branches: [master]
  schedule:
    - cron: '0 4 * * *'
```

Then a change in `objectionary/home` or in Maven Central shows up as a red `master` the
next morning, on its own, and not as somebody's unrelated pull request.

2. A merge should not go in on a stale tick. GitHub's "Require branches to be up to date
before merging" re-runs the checks against the current `master` before the merge button
works, which also covers the case where two pull requests are fine apart and broken
together.

Worth considering as well: the run pulls objects from a pinned tag but the set of objects
it pulls is decided at run time, so a pinned tag is not by itself enough to make the job
reproducible. A check that the pulled set is closed under the tag, or a lock file listing
what was pulled, would make the failure show up as a diff rather than as a 404.

morphqdd:
`itest` pulls objects from `objectionary/home` while it runs and `grunt` resolves
`eo-maven-plugin` from Maven Central, so both can turn red without anything here
changing. They ran only on push and on pull request, so nobody learned about it until the
next contributor opened a pull request that had nothing to do with it.

Both now also run at 04:00 UTC, so such a break shows up as a red `master` on its own.

The other half of the report, requiring a branch to be up to date before merging, is a
repository setting and cannot be done from here.

Closes #1290

### DIFF
--- a/.github/workflows/grunt.yml
+++ b/.github/workflows/grunt.yml
@@ -10,6 +10,8 @@ name: grunt
   pull_request:
     branches:
       - master
+  schedule:
+    - cron: '0 4 * * *'
 concurrency:
   group: grunt-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/itest.yml
+++ b/.github/workflows/itest.yml
@@ -10,6 +10,8 @@ name: itest
   pull_request:
     branches:
       - master
+  schedule:
+    - cron: '0 4 * * *'
 concurrency:
   group: itest-${{ github.ref }}
   cancel-in-progress: true


yegor256:
Thanks for the contribution, merged!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6804](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1300, opened 2026-09-07
#1294: Reject missing source directories before register

### LINKED ISSUE #1294: `eoc register` incorrectly reports success and exits with 0 when source directory is missing
## What happens

`eoc register` pointed at a sources directory that does not exist reports success, exits `0`, and writes no catalog:

```
$ eoc --alone --batch -s ./nosrc -t ./t register
EO objects registered in C:\…\t\eo-foreign.json in 8s
$ echo $?
0
$ ls t/eo-foreign.json
(nothing — the file was never created)
```

The very next command proves the file is absent:

```
$ eoc --alone --batch -s ./nosrc -t ./t foreign
ENOENT: no such file or directory, open 'C:\…\t\eo-foreign.json'
```

So the message names an artefact that does not exist, and a script that checks the exit code sees a successful registration of nothing.

## Root cause

`src/commands/register.js:16-22` prints its line as soon as Maven returns, without asking whether anything was registered:

```js
const foreign = path.resolve(opts.target, 'eo-foreign.json');
return elapsed(async (tracked) => {
  const r = await mvnw(goals().concat(flags(opts)), opts.target, opts.batch);
  tracked.print(`EO objects registered in ${rel(foreign)}`);
  return r;
});
```

`mvnw` resolves with the arguments whenever Maven exits `0`, and `eo:register` over a directory with no `.eo` files inside it is not an error for the plugin — it simply has nothing to write, so it writes nothing.

`flags()` in `src/mvnw.js` does check that `--sources` was *given* ("Sources directory is not specified…"), but not that it exists, so a typo in the path takes this path rather than the diagnostic one.

## Suggested fix

Two small guards, either of which removes the false report:

```js
const sources = path.resolve(opts.sources);
if (!fs.existsSync(sources)) {
  throw new Error(`The sources directory ${rel(sources)} does not exist`);
}
```

in `mvnw.flags()`, where the other `--sources` check already lives, and after the Maven call:

```js
if (!fs.existsSync(foreign)) {
  tracked.print(`No EO sources found in ${rel(sources)}, nothing registered`);
} else {
  tracked.print(`EO objects registered in ${rel(foreign)}`);
}
```

so that an empty but existing directory is reported honestly too.

The neighbouring commands have the same shape of report — `eoc print` (#1166) and `eoc docs` (#1241) both announce completion over inputs that were never produced — so a single "did this actually produce anything?" check is worth applying across the three.

gemshrine:
Fixes #1294.

`eoc register` accepted a `--sources` path that did not exist. Maven exits
successfully when there are no inputs to process, so the command printed
`EO objects registered in .../eo-foreign.json` and returned exit code `0` even
though no catalog was created. This made a misspelled path look like a
successful registration and caused the following `foreign` command to fail
with an unrelated missing-file error.

`mvnw.flags()` now resolves the source path and verifies that it exists before
Maven is started. A missing directory raises a clear error, preventing the
success message and non-zero completion from being suppressed. Existing source
directories and empty-directory behavior are unchanged; this check only
rejects an invalid path.

A regression test exercises the flag builder with a guaranteed-absent temporary
path and checks the diagnostic. Existing flag tests now use the repository
directory as their valid source fixture.

Verification:

- `npx mocha test/test_mvnw.js --timeout 1200000` — 15 tests, 0 failures.

### DIFF
--- a/src/mvnw.js
+++ b/src/mvnw.js
@@ -50,6 +50,9 @@ module.exports.flags = function(opts) {
     throw new Error('Target directory is not specified. Please provide it with --target option.');
   }
   const sources = path.resolve(opts.sources);
+  if (!fs.existsSync(sources)) {
+    throw new Error(`Sources directory ${rel(sources)} does not exist.`);
+  }
   console.debug('Sources in %s', rel(sources));
   const target = path.resolve(opts.target);
   console.debug('Target in %s', rel(target));
--- a/test/commands/java/test_pipeline.js
+++ b/test/commands/java/test_pipeline.js
@@ -16,7 +16,7 @@ describe('java/pipeline', () => {
     const maven = function maven(command) {
       calls.push(command);
     }
-    const opts = {sources: 'sources-dir', target: 'target-dir'};
+    const opts = {sources: '.', target: 'target-dir'};
     await pipeline(coms, ['register', 'assemble'], opts, maven)
     assert.deepStrictEqual(
       calls[0],
@@ -50,7 +50,7 @@ describe('java/pipeline', () => {
       calls.push(command);
     }
     const opts = {
-      sources: 'lint-sources-dir',
+      sources: '.',
       target: 'lint-target-dir',
       newParser: true,
       easy: true,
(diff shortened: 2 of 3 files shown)


yegor256:
CI is green and the fix addresses #1294, but I'm not confident enough to merge it without a human look, so I'm assigning it for manual review.

The concern is scope. #1294 reports that `eoc register` reports success on a missing source directory, but the guard is placed in `mvnw.flags()`, which every Maven-driven command goes through. After this change any command that builds flags will refuse to run when the sources directory is absent, not just `register`. For commands that do not read sources at all — `audit` and `clean` are the ones I'd check first — that turns a previously working invocation into a hard error in a project that simply has no sources directory yet.

Either the guard belongs closer to `register`, or the widening is intentional and worth stating in the description so it is a deliberate behaviour change rather than a side effect. The rewriting of the existing fixtures from named paths to `.` is a symptom of the same reach: those tests were not about the filesystem, and now they depend on it.

---

PR #1301, opened 2026-09-07
#1293: Explain how to create a missing foreign catalog

### LINKED ISSUE #1293: `eoc foreign` command on an unregistered project returns raw `ENOENT` error instead of instructing to run `eoc register`
## What happens

`eoc foreign` on a project that has not been registered yet prints a raw file-system error, naming a path the user never typed and saying nothing about what to do:

```
$ eoc --alone --batch -s src -t .eoc foreign
ENOENT: no such file or directory, open 'C:\…\.eoc\eo-foreign.json'
```

The catalog is produced by `eoc register` (`src/commands/register.js` prints "EO objects registered in …/eo-foreign.json"), so the file is simply not there yet — an ordinary situation on a fresh checkout, and the one a new user hits first when exploring the commands.

## Root cause

`src/commands/foreign.js:57-58` reads the file with no check:

```js
const file = path.resolve(opts.target, 'eo-foreign.json');
const all = JSON.parse(firstJsonArray(fs.readFileSync(file, 'utf8')));
```

`readFileSync` throws `ENOENT`, the message is whatever Node produced, and `eoc.js` catches it at the top level to print `e.message` — so the exception's text becomes the whole diagnostic.

The same file already shows what a considered failure looks like: `firstJsonArray` throws `SyntaxError('No complete JSON array found in foreign catalog')` when the content is unusable, which names the artefact rather than the syscall.

## Suggested fix

Say what is missing and how it is made:

```js
if (!fs.existsSync(file)) {
  throw new Error(
    `There is no ${rel(file)} yet, run "eoc register" first to create it`
  );
}
```

Two neighbours are worth the same treatment, since both read the same catalog and fail the same way on a fresh target: `src/commands/foreign.js` is the only one that reads it directly today, but `eoc print` and `eoc docs` read directories the pipeline may not have produced and report success instead (#1166, #1241) — the three together are the "target directory is empty" case, and a single guard shape would cover them.

gemshrine:
Fixes #1293.

`eoc foreign` attempted to read `eo-foreign.json` unconditionally. On a fresh
or unregistered target this produced Node's raw `ENOENT` message, exposing an
implementation path and leaving the user without the next action. The catalog
is normally created by `eoc register`, so a missing file is an expected setup
state rather than an unexpected system failure.

The command now checks for the catalog before reading it and raises a concise
error that names the relative file and tells the user to run `eoc register`.
Valid catalogs still use the existing parser and report unchanged; malformed
catalog contents continue to receive the existing `No complete JSON array`
diagnostic.

A regression test invokes the command with a missing target and verifies the
actionable message. The existing integration test still confirms inspection of
a registered catalog.

Verification:

- `npx mocha test/commands/test_foreign.js --timeout 1200000` — 2 tests, 0 failures.

### DIFF
--- a/src/commands/foreign.js
+++ b/src/commands/foreign.js
@@ -53,6 +53,11 @@ function firstJsonArray(text) {
  */
 module.exports = function(opts) {
   const file = path.resolve(opts.target, 'eo-foreign.json');
+  if (!fs.existsSync(file)) {
+    throw new Error(
+      `There is no ${rel(file)} yet, run "eoc register" first to create it`
+    );
+  }
   const all = JSON.parse(firstJsonArray(fs.readFileSync(file, 'utf8')));
   console.info('There are %d objects in %s:', all.length, rel(file));
   all.forEach((obj) => {
--- a/test/commands/test_foreign.js
+++ b/test/commands/test_foreign.js
@@ -7,9 +7,17 @@ const assert = require('assert');
 const fs = require('fs');
 const path = require('path');
 const {runSync, assertFilesExist, parserVersion, homeTag, weAreOnline} = require('../helpers');
+const foreign = require('../../src/commands/foreign');
 
 describe('foreign', () => {
   before(weAreOnline);
+  it('explains how to create a missing foreign catalog', () => {
+    const target = path.join('temp', `test-foreign-missing-${Date.now()}`);
+    assert.throws(
+      () => foreign({target}),
+      /There is no .*eo-foreign\.json yet, run "eoc register" first to create it/
+    );
+  });
   it('inspects foreign objects and prints a report', (done) => {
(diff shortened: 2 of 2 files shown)


yegor256:
Merged. Thanks — a missing catalog now tells the user to run `eoc register` instead of leaking `ENOENT`.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4030](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1303, opened 2026-09-07
#1291: Reject inherited names as language values

### LINKED ISSUE #1291: `--language constructor` results in misleading diagnostic message due to platform lookup validation issue
## What happens

`--language` is validated against a plain object, so a value that names a member of `Object.prototype` passes the check and comes back as that member instead of a platform name. The mistake then surfaces from a different place, in a message that names a JavaScript internal rather than what the user typed.

```
$ eoc --language constructor clean
Unknown platform function Object() { [native code] }

$ eoc --language __proto__ clean
Unknown platform [object Object]
```

The control is the ordinary case, which is reported the way it should be, by commander, at parse time, naming the flag:

```
$ eoc --language nope clean
error: option '-l, --language <name>' argument 'nope' is invalid. Unknown platform nope
```

All three exit `1`, so the damage is confined to the diagnostic, but the two above are unreadable: nothing in them says which option was wrong, and `function Object() { [native code] }` is not a string the user can act on.

## Root cause

`src/eoc.js:80-83` builds the lookup as an object literal:

```js
const platforms = {};
for (const [alias, canonical] of Object.entries(language)) {
  platforms[alias] = canonical;
  platforms[canonical.toLowerCase()] = canonical;
}
```

and `canonicalLanguage` decides on `undefined`:

```js
const canonical = platforms[String(value).toLowerCase()];
if (canonical === undefined) {
  throw new InvalidArgumentError(`Unknown platform ${value}`);
}
return canonical;
```

`{}` inherits from `Object.prototype`, so `platforms['constructor']` and `platforms['__proto__']` are not `undefined` — they are the inherited members. The guard passes and a function (or the prototype object) is returned as if it were `'Java'` or `'JavaScript'`.

Checked against `f900dac`:

| `--language` | `canonicalLanguage(value)` |
| --- | --- |
| `js` | `'JavaScript'` |
| `nope` | throws `InvalidArgumentError` |
| `toString` | throws — the input is lower-cased first, and `tostring` is not a member |
| `constructor` | **returns** `function Object() { [native code] }` |
| `__proto__` | **returns** `[object Object]` |

The value then travels into `select()` at `src/eoc.js:523`, where `registry[canonicalLanguage(lang)]` looks up a key that does not exist, so `coms()` yields `undefined` and the failure is reported by the second call to `canonicalLanguage`, interpolating the function into the message.

## Suggested fix

Give the lookup no prototype, so an inherited member cannot be mistaken for an entry:

```js
const platforms = Object.create(null);
```

or ask the object rather than the value:

```js
if (!Object.hasOwn(platforms, String(value).toLowerCase())) {
  throw new InvalidArgumentError(`Unknown platform ${value}`);
}
```

Either one keeps `--language constructor` on the same path as `--language nope`, which is where it belongs.

gemshrine:
Fixes #1291.

The `--language` aliases were stored in a normal object. Since normal objects
inherit from `Object.prototype`, values such as `constructor` and `__proto__`
looked like successful platform lookups and returned a function or prototype
object. The later command selection then failed with a misleading diagnostic
such as `Unknown platform function Object() { [native code] }` instead of
identifying the supplied option value.

The platform lookup is now a prototype-free map. Only explicitly registered
aliases can resolve, so inherited property names follow the same
`InvalidArgumentError` path as any other unknown language while valid aliases
and case-insensitive canonical names remain unchanged.

Regression coverage checks both problematic inherited names and retains the
existing canonicalization and command-selection cases.

Verification:

- `npx mocha test/test_eoc.js --timeout 1200000` — 33 tests, 0 failures.

### DIFF
--- a/src/eoc.js
+++ b/src/eoc.js
@@ -77,7 +77,7 @@ const pipelines = {
  * lower-cased (e.g. `javascript`); `canonicalLanguage()` lower-cases the
  * input, so mixed-case spellings such as `JavaScript` are matched too.
  */
-const platforms = {};
+const platforms = Object.create(null);
 for (const [alias, canonical] of Object.entries(language)) {
   platforms[alias] = canonical;
   platforms[canonical.toLowerCase()] = canonical;
--- a/test/test_eoc.js
+++ b/test/test_eoc.js
@@ -151,6 +151,16 @@ describe('canonicalLanguage', () => {
     );
     done();
   });
+  it('rejects inherited object property names as platforms', () => {
+    assert.throws(
+      () => canonicalLanguage('constructor'),
+      /Unknown platform constructor/
+    );
+    assert.throws(
+      () => canonicalLanguage('__proto__'),
+      /Unknown platform __proto__/
+    );
+  });
 });
 
 describe('select', () => {


yegor256:
Merged. Thanks — the prototype-free lookup is the right fix, and the regression covers both inherited names.

---

0crat:
@gemshrine Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+4054](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1306, opened 2026-09-08
Remove the no-op audit command

### LINKED ISSUE #1264: `eoc audit` command doesn't inspect packages but merely prints Maven version
## What happens

`eoc audit` inspects nothing. It prints the version of Maven and returns.

`README.md:119` and the `--help` output both say:

```
* `audit` Inspect all packages and report their status
```

`src/commands/audit.js` in full:

```js
const {mvnw} = require('../mvnw');

  return mvnw(['--version'], null, opts.batch);
};
```

`mvnw(['--version'])` runs `mvnw --version`, which prints the Maven banner and exits. No package is read, no status is computed, nothing is reported. A user who runs `eoc audit` to find out whether their dependencies are in order gets:

```
Apache Maven 3.9.16 ...
Java version: ...
```

and no indication that the command did not do what it offered to do.

## Root cause

The test certifies the current behaviour rather than the promised one. `test/commands/test_audit.js`:

```js
describe('audit', () => {
  it('audits all packages', (done) => {
    const stdout = runSync(['audit']);
    assert(stdout.includes('Apache Maven'), stdout);
    done();
  });
});
```

The name says the command audits all packages; the assertion says the output contains `Apache Maven`, which is true of `mvnw --version` and would stay true if the audit were removed entirely. So the gap is held in place from both sides: the command does nothing, and the test agrees that doing nothing is correct.

## Suggested fix

Either implement the command or stop advertising it.

The data it would need is already gathered. `eo-foreign.json` under the target directory carries every foreign object and its state, and `src/commands/foreign.js` already reads it. An audit could report per-package counts of what is resolved, what is missing, and what is stale, on top of that same file, with the Maven `--version` line dropped as unrelated.

If nobody needs it, removing the command, its README line and its test is honest and smaller. What should not stay is the present arrangement, where the help text promises a report the code was never written to produce.

Either way the test needs an assertion about the audit rather than about Maven's banner, so that the next change to this command is caught by it.

gemshrine:
## What changed

Remove the `audit` command, its misleading help entry, and the test that only asserted Maven's version banner. The command never inspected packages; it only ran `mvnw --version`, so keeping it advertised made a successful no-op look like an audit report.

The existing `foreign` command remains available for inspecting the generated foreign-object catalog.

Closes #1264.


yegor256:
CI is green and the diff is clean, but I don't want to merge this without a maintainer's call: #1264 can be resolved two ways, and #1255 already proposes the other one (make `audit` actually report the packages the build depends on). Removing a documented command is a user-facing break, so the two PRs should be decided together rather than merged first-come. Assigning to @yegor256.

---

PR #1322, opened 2026-09-10
#1309: Await Java dataization process

### LINKED ISSUE #1309: `dataize` function in `src/commands/java/dataize.js` returns `undefined`, not providing promise for JVM completion
## The bug
`src/commands/java/dataize.js` starts the JVM and attaches a `close` handler, but the exported function returns `undefined` instead of the child completion promise. The CLI and any library caller therefore continue as soon as `spawn()` returns; `await coms().dataize(...)` does not wait for dataization or propagate a successful exit.

## Reproduction
Call the command with a stub runner that returns an EventEmitter and check the result:

```js
const result = dataize('main.foo', [], opts, () => true, runner)
// result === undefined, even though the JVM has not emitted `close` yet
```

The same behavior occurs in `src/eoc.js`, which explicitly awaits `coms().dataize()`.

## Expected
Return a Promise that resolves after the JVM closes successfully and rejects on a non-zero exit (without terminating the hosting process), so callers can reliably await the command and handle failures.

zaguzovmaksim0-hue:
Closes #1309.

Java `dataize()` now returns a Promise tied to the spawned JVM instead of returning `undefined` immediately. The Promise stays pending until the child closes, resolves on exit code 0, and rejects on either a non-zero exit code or a child-process `error`.

This makes the existing `await coms().dataize(...)` in `eoc.js` actually wait for JVM completion and lets library callers handle failures without `dataize()` terminating the hosting process itself.

Added focused tests with an `EventEmitter` child stub proving that the Promise does not settle before `close`, and that both non-zero close and process errors reject.

Validation:
- `npx eslint src/commands/java/dataize.js test/commands/test_dataize.js`
- `npx mocha test/commands/test_dataize.js --grep 'dataize/java' --timeout 20000`

Both pass locally.

### DIFF
--- a/src/commands/java/dataize.js
+++ b/src/commands/java/dataize.js
@@ -14,6 +14,7 @@ const {verifyJavac} = require('../../jdk');
  * @param {Object} opts - All options
  * @param {Function} [exec] - Optional command runner for the JDK check
  * @param {Function} [runner] - Optional Java process runner
+ * @return {Promise} Resolves when the JVM exits successfully
  */
 module.exports = function(obj, args, opts, exec, runner = spawn) {
   verifyJavac(exec);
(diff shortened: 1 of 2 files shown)


yegor256:
Merged, thanks! Both call sites in `src/eoc.js` already `await coms().dataize(...)`, so returning a real promise is what makes that `await` mean anything. The user-visible behaviour on failure is unchanged too — the rejection reaches the `parseAsync` handler, which prints the message and exits 1, exactly as the old `process.exit(1)` did. The test that asserts the promise is still pending before `close` fires is the one that actually pins the bug.

---

0crat:
@zaguzovmaksim0-hue Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+116](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1324, opened 2026-09-10
#1314: Fail when Codecov upload fails

### LINKED ISSUE #1314: Codecov upload failures are explicitly ignored in CI
## The bug
The Codecov step in `.github/workflows/codecov.yml` sets `fail_ci_if_error: false`. If the coverage artifact is malformed, the token is rejected, or Codecov is unavailable, the workflow still succeeds and reports green despite failing to publish its only coverage result.

## Expected
Treat a failed coverage upload as a failed coverage job (or add a separate explicit verification step), so a green check means the report was actually accepted.

zaguzovmaksim0-hue:
Closes #1314.

The Codecov action now uses `fail_ci_if_error: true`, so the trusted master coverage job fails when Codecov rejects or cannot publish the coverage report instead of silently staying green.

The upload step is already restricted to `github.ref == 'refs/heads/master'`, where the repository token is available, so this does not make fork pull requests depend on a secret.

The workflow YAML parses successfully locally.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -39,5 +39,5 @@ jobs:
           files: ./coverage/lcov.info
           flags: unittests
           name: codecov-umbrella
-          fail_ci_if_error: false
+          fail_ci_if_error: true
         if: github.ref == 'refs/heads/master'


zaguzovmaksim0-hue:
Closing this PR to comply with the maintainer policy: issue #1314 is currently unlabeled. I will not submit work against unlabeled issues; if the issue is accepted/labeled later, the work can be reconsidered then.

PR #1329, opened 2026-09-10
#1316: Pin PDD action to immutable commit

### LINKED ISSUE #1316: PDD workflow runs the mutable master branch of pdd-action
## The bug
`.github/workflows/pdd.yml` uses `volodya-lombrozo/pdd-action@master`. The action code can change independently of this repository, altering or breaking validation for an unchanged EOC commit.

## Expected
Pin the action to an immutable commit SHA and document its release, keeping the check reproducible.

zaguzovmaksim0-hue:
Closes #1316.

Pins `volodya-lombrozo/pdd-action` to immutable commit `69842b56627431c5f232f5ea2dc2fca82f409c54`, which is the current upstream `master` commit (2023-12-18).

The upstream action has no GitHub releases and no tags, so there is no release tag that can be documented without inventing one. Pinning the exact commit currently referenced by `master` preserves today's behavior while making the EOC check reproducible; the inline comment records the upstream date/version identity.

The workflow YAML parses successfully locally.

### DIFF
--- a/.github/workflows/pdd.yml
+++ b/.github/workflows/pdd.yml
@@ -16,4 +16,4 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v7
-      - uses: volodya-lombrozo/pdd-action@master
+      - uses: volodya-lombrozo/pdd-action@69842b56627431c5f232f5ea2dc2fca82f409c54 # master 2023-12-18


zaguzovmaksim0-hue:
Closing this PR to comply with the maintainer policy: issue #1316 is currently unlabeled. I will not submit work against unlabeled issues; if the issue is accepted/labeled later, the work can be reconsidered then.

PR #1331, opened 2026-09-10
#1310: Isolate Maven progress state per run

### LINKED ISSUE #1310: `mvnw()` calls overwrite shared progress state when run concurrently
## The bug
`src/mvnw.js` stores the current build's `target`, `phase`, `beginning`, and `running` values in module-level variables (`src/mvnw.js:37-41`). Every non-batch `mvnw()` invocation overwrites those variables, while `start()`, `print()`, and `stop()` read them asynchronously.

## Reproduction
Start two non-batch calls before either child emits `close`, using two targets and different goals. The progress line for the first process reports the second process's phase/target; when either process closes, `stop()` sets the shared `running` flag to false and progress reporting for the other process stops prematurely.

## Expected
Progress state should belong to each invocation (or concurrent calls should be serialized), so one Maven process cannot cancel or relabel another process's status output.

zaguzovmaksim0-hue:
Closes #1310.

Moves Maven progress state from module-level variables into an object created for each `mvnw()` invocation. `start`, `print`, and `stop` now operate on that invocation's state, so concurrent runs cannot overwrite each other's phase/target or stop each other's ticker.

The optional process runner parameter is only for deterministic testing and defaults to the existing `spawn`, preserving callers.

Added a regression test with two independent `EventEmitter` child processes and targets containing different file counts. It verifies both tickers continue with their own labels, then closes the first process and proves the second ticker keeps running.

Validation:
- `npx eslint src/mvnw.js test/test_mvnw.js`
- focused Mocha regression test
- `git diff --check`

All pass locally.

### DIFF
--- a/src/mvnw.js
+++ b/src/mvnw.js
@@ -32,11 +32,6 @@ function shell() {
   }
 }
 
-let beginning,
-  phase = 'unknown',
-  running = false,
-  target;
-
 /**
  * Prepare options for Maven.
  * @param {Object} opts - Opts provided to the "eoc"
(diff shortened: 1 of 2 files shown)


yegor256:
The production change is right and I would happily take it: moving `beginning`, `phase`, `running` and `target` off the module and into a per-call `progress` object is exactly what #1310 asks for, and threading `runner` in follows the pattern `commands/java/dataize.js` already uses. CI is green.

What stops me merging is the test, so this is for @yegor256 to weigh. `keeps concurrent progress state independent` asserts `firstTicks >= 2` after sleeping 1100ms, and the ticker prints once immediately and then once per 1000ms — so the second tick has about 100ms of slack. On a loaded runner, or on the Windows job where this suite is slowest, one tick is enough to make it red, and it would read as an unrelated failure to whoever hits it. The later assertion has the same shape.

The same property can be pinned without the clock, for example by exposing the tick as a function the test calls twice and asserting that two `mvnw` calls render their own `phase` and `target`. If you would rather keep the timing test, raising the waits well clear of the interval would at least make it hard to lose a tick.

---

PR #1332, opened 2026-09-10
#1147: Link generated docs stylesheet relatively

### LINKED ISSUE #1147: Generated docs (eoc docs) with absolute css path fail to load stylesheet on Windows and expose local paths

zaguzovmaksim0-hue:
Closes #1147.

Generated HTML now receives a stylesheet href relative to the HTML file itself instead of the absolute local path to `styles.css`. The helper also normalizes the result to `/`, so generated links remain valid URLs on Windows.

This handles each output depth independently:
- nested object pages link back through the required `../` segments;
- package pages and `packages.html`, which live beside `styles.css`, use `styles.css` directly.

The existing docs regression now checks the actual hrefs for both nested and root-level generated pages, so an absolute filesystem path or a depth-insensitive link will fail the test.

Validation on current master:
- `npx eslint src/commands/docs.js test/commands/test_docs.js`
- `npx mocha test/commands/test_docs.js --grep 'generates HTML files for files and packages' --timeout 120000`
- `git diff --check`

All pass locally. This is a clean current-master replacement for the closed/stale #1152; it does not reuse that PR's malformed diff markers.

### DIFF
--- a/src/commands/docs.js
+++ b/src/commands/docs.js
@@ -117,6 +117,16 @@ function wrapHtml(name, html, css) {
 }
 
 /**
+ * Create a browser-safe relative link from one generated file to another.
+ * @param {String} source - Generated HTML file path
+ * @param {String} target - Linked asset path
+ * @return {String} Relative URL using forward slashes
+ */
+function relativeHref(source, target) {
+  return path.relative(path.dirname(source), target).split(path.sep).join('/');
+}
+
+/**
  * Command to generate documentation.
  * @param {Hash} opts - All options
  * @return {Promise<String>} Resolves to the message reporting the summary path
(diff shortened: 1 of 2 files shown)


yegor256:
This fixes #1147, and #1277 is already open for #1262, which reports the same defect: `eoc docs` links the stylesheet by absolute filesystem path, so it never loads on Windows and breaks when the docs move. #1277 is under review by @yegor256. Two changes to the same link cannot both land, so I am not merging this one; over to @yegor256 to pick, and the duplicate issue can be closed against whichever survives.

Same situation applies to #1334 and #1281, which both fix #1267.

---

PR #1333, opened 2026-09-10
#1308: Run fallback Maven from caller directory

### LINKED ISSUE #1308: `mvnw` fallback incorrectly executes system Maven from wrong directory
When the bundled wrapper is absent, `src/mvnw.js` falls back to the system `mvn` (`src/mvnw.js:97-103`) but keeps `cwd: home` in the spawn options (`src/mvnw.js:113-120`). `home` is always `<eoc>/mvnw`, not the user project or the target passed to `mvnw()`.

An installed/distributed eoc without its repository wrapper therefore runs system Maven from the eoc package's `mvnw` directory. Maven cannot see the caller's `pom.xml` there and fails with a missing-project error (or operates on the wrong project), while the same command works when invoked from the project directory.

Use the caller's project/target directory (or preserve the caller's current working directory) for the fallback, while retaining the wrapper directory only for the bundled script. Add a test that forces the wrapper-missing branch and asserts the spawn `cwd`.

zaguzovmaksim0-hue:
Closes #1308.

When the bundled wrapper is unavailable, system `mvn` now keeps the caller's current working directory instead of inheriting the eoc package's internal `mvnw/` directory. The bundled wrapper still runs from its own directory as before.

The executable/cwd selection is factored into `maven()` so both branches are regression-tested. The tests verify that a missing wrapper selects `mvn` with `process.cwd()`, while an existing bundled wrapper keeps its wrapper directory.

Validation performed locally:
- `npx mocha test/test_mvnw.js --timeout 1200000` — passed;
- `npx eslint src/mvnw.js test/test_mvnw.js` — passed;
- `git diff --check` — clean.

### DIFF
--- a/src/mvnw.js
+++ b/src/mvnw.js
@@ -83,6 +83,22 @@ module.exports.flags = function(opts) {
 };
 
 /**
+ * Resolve Maven executable and its working directory.
+ * @param {String} home - Directory containing the bundled wrapper
+ * @return {Object} Maven binary and working directory
+ */
+module.exports.maven = function(home) {
+  let bin = path.resolve(home, 'mvnw') + (process.platform === 'win32' ? '.cmd' : '');
+  let cwd = home;
+  if (!fs.existsSync(bin)) {
+    console.warn(colors.yellow(`Warning: mvnw not found at ${bin}, falling back to system "mvn"`));
+    bin = 'mvn';
+    cwd = process.cwd();
+  }
+  return {bin, cwd};
+};
+
+/**
  * Run mvnw with provided commands.
  * @param {Array.<String>} args - All arguments to pass to it
  * @param {String} [tgt] - Path to the target directory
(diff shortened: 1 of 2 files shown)


yegor256:
I do not think this is the right change, so I am not merging it and am handing it to @yegor256 with the reasoning.

The premise of #1308 is that `home` is "not the user project", so Maven run from there "cannot see the caller's `pom.xml`". But eoc never asks Maven to build the caller's project. `mvnw/pom.xml` is eoc's own POM — `org.eolang:eoc:0.0.0`, with `eo-runtime` as its dependency and `${eo.generatedDir}` as its source directory — and that is the project every `mvnw()` invocation is meant to build. A user of eoc has no `pom.xml` of their own at all.

So `cwd: home` is not a leftover from the bundled branch; it is the only directory where the POM being built lives. Pointing the fallback at `process.cwd()` makes system `mvn` look for a project in the user's working directory and fail with a missing-POM error — the very failure the issue is trying to remove, just moved to the branch this change touches.

Worth noting too that the branch is close to unreachable: `.npmignore` excludes `test/`, `itest/`, `.github/` and friends, but not `mvnw/`, so `mvnw/mvnw`, `mvnw/mvnw.cmd` and `mvnw/pom.xml` all ship inside the published package.

If the warning branch is dead weight, removing it is a reasonable change. Redirecting its working directory is not.

---

PR #1343, opened 2026-09-16
Create missing generate-comments output directories

### LINKED ISSUE #1342: `generate_comments` command does not create or warn about non-existent output parent directory
eoc generate_comments fails when the output parent directory does not exist.

### Steps to reproduce

1. Create a valid source file and prompt template.
2. Run generate_comments with an output path such as tmp/missing/out.json, where tmp/missing does not exist.
3. Use the placeholder provider to avoid any external model dependency.

### Actual result

The command generates the result, then writeFileSync fails because the parent directory is absent.

### Expected result

The command should create the output parent directory or fail with a clear message before doing the generation work.

The affected code is src/commands/generate_comments.js, where the output is written directly.

gemshrine:
Closes #1342

### Problem

`generate_comments` generated all responses successfully and then called `writeFileSync` directly. When the requested output path was inside a missing directory, the command failed after doing the generation work with a raw filesystem error.

### Solution

Create the parent directory recursively immediately before writing the generated JSON. Existing output paths keep their behavior, while nested output paths now work on a clean target.

### Verification

- Added a regression test for an output path whose parent directory does not exist.
- `npm test -- --grep "creates a missing output directory"` could not run in this checkout because dependencies are not installed; CI will run the full Mocha suite.
- `git diff --check`


yegor256:
Merged, thanks!

---

PR #1344, opened 2026-09-16
Preserve custom docs stylesheets

### LINKED ISSUE #1340: `eoc docs` command overwrites custom styles in `styles.css` file on each execution
## What happens

Every invocation of `eoc docs` unconditionally executes `fs.writeFileSync(css, '')` for `<target>/docs/styles.css`. If a user has customized or post-processed that stylesheet, the next documentation build silently erases it and all generated pages lose the custom rules.

## How to reproduce

1. Run `eoc docs` once so `<target>/docs/styles.css` exists.
2. Put any non-empty CSS in that file (or let a downstream documentation step populate it).
3. Run `eoc docs` again.
4. The file is truncated to zero bytes before the HTML pages are written.

## Expected result

The command should either generate a deterministic stylesheet from a source asset, preserve an existing stylesheet, or write a new generated file without destroying user content. A rebuild of documentation should not erase a file it does not own.

## Evidence

`src/commands/docs.js` creates the output directory and then calls `fs.writeFileSync(css, '')` unconditionally before processing XMIR files. There is no check for an existing file and no bundled stylesheet copied in its place.

gemshrine:
Closes #1340

### Problem
Each `eoc docs` run truncates `docs/styles.css` before generating the documentation. A project that customizes the generated stylesheet therefore loses those changes on every subsequent build.

### Solution
Create `styles.css` only when it does not exist. Existing stylesheet contents are left untouched while the generated HTML continues to reference the same file.

### Verification
- Added a regression test that runs documentation generation twice and verifies custom CSS survives the second run.
- `git diff --check` passes.
- The repository dependencies are not installed in this checkout, so the JavaScript test suite could not be run locally; CI should run the complete test suite.


yegor256:
Merged, thanks!

---

PR #1347, opened 2026-09-16
Stop waiting when inspect server exits

### LINKED ISSUE #1337: `eoc inspect` unresponsive for full 60-second timeout after premature Java server exit
## What happens

`eoc inspect` waits for the complete one-minute polling timeout even when the inspection JVM has already exited. A failed server startup is therefore reported as “has not opened port … in a minute” only after 60 seconds, instead of returning the JVM's actual failure immediately.

## How to reproduce

1. Run `eoc inspect --alone` with a target whose inspection server cannot start (for example, an invalid `--port` or a target with a broken inspection classpath).
2. The spawned `java` process exits before opening the port.
3. Observe that `src/commands/java/inspect.js` keeps calling `ask()` until its deadline; it never observes the child's `close`/`error` event.

The `finally` block kills the child only after `ask()` rejects, so the early process exit is invisible to the caller and the command blocks for up to 60 seconds.

## Expected result

When the child emits `error` or exits before the port is ready, `eoc inspect` should reject immediately and include the process exit status/stderr. The one-minute timeout should be reserved for a live server that is still starting.

## Evidence

`module.exports` starts `runner('java', ...)` and then calls `await ask(...)`; `ask()` only polls HTTP and has no connection to the child process lifecycle. No listener handles `close` or `error` before the deadline.

gemshrine:
Closes #1337

### Problem
`eoc inspect` waits for the full polling deadline when its Java server exits before opening the port. The command then reports a misleading timeout instead of the actual startup failure.

### Solution
Race the port polling request against the inspection process lifecycle. Early `error` and `close` events now reject immediately with a diagnostic that includes the startup error or exit code. Listeners are removed during cleanup before the child is killed.

### Verification
- Added a regression test that emits an early process exit and verifies immediate rejection with the exit status.
- `node --check` passes for the changed JavaScript files.
- The repository dependencies are not installed in this checkout, so the full test suite could not be run locally; CI should run it.


yegor256:
Merged, thanks!

---

PR #1348, opened 2026-09-16
Traverse symlinked source directories

### LINKED ISSUE #1341: `eoc normalize` and `docs` commands overlook files in symlinked source directories
## What happens

The recursive file helpers do not descend into symbolic links to directories. As a result, source files stored below a symlinked directory are silently omitted from `normalize` and other commands that use `findFiles` or `copyDir`.

## How to reproduce

1. Create a project source tree containing `real/main.eo`.
2. Create `sources/link` as a symbolic link to `real`.
3. Run the command that scans `sources` (for example, `eoc normalize`).
4. Inspect the files selected for parsing or copied to `before-normalize`.

`src/files.js` calls `fs.readdirSync(..., {withFileTypes: true})` and recurses only when `entry.isDirectory()` is true. For a directory symlink `Dirent#isDirectory()` is false, so `real/main.eo` is never visited.

## Expected result

A symlinked source directory should either be traversed with documented cycle protection or rejected with an explicit diagnostic. It should not make valid source files disappear from the generated output.

## Evidence

The omission happens before Maven or phino runs, and no warning is emitted. Projects using linked source trees therefore receive incomplete output while the command still reports success.

gemshrine:
Closes #1341

### Problem
The recursive file helpers relied on `Dirent.isDirectory()`, so directory symlinks were treated as files and their source files were silently skipped. Following a symlink loop could also recurse forever if traversal were enabled without protection.

### Solution
Resolve symlink targets with `statSync` when deciding whether an entry is a directory, and track visited real paths during each traversal. `findFiles` and `copyDir` now include files below linked directories while stopping at cycles.

### Verification
- Added a regression test covering a linked source directory and a loop back to its parent.
- `node --check` passes for the changed JavaScript files.
- `git diff --check` passes.
- The repository dependencies are not installed in this checkout, so the full test suite could not be run locally; CI should run it.


yegor256:
Merged, thanks!

---

PR #1361, opened 2026-09-20
Reject mvnw spawn errors

### LINKED ISSUE #1360: `mvnw` Promise remains pending instead of rejecting on child process `error` event
Steps to reproduce

1. Run an eoc command when the wrapper executable exists but cannot be spawned, for example because it is not executable.
2. Observe the returned Promise.

Actual result

`spawn` emits an `error` event, but `mvnw` registers only a `close` handler. The error is uncaught or the Promise remains pending instead of rejecting with the spawn failure.

Expected result

The command should reject immediately on the child `error` event and stop any progress reporting.

Evidence

`src/mvnw.js` creates the child process and listens for `close` only. There is no `result.on(error, reject)` handler.

gemshrine:
Fixes #1360


yegor256:
CI is green here, but #1252 is already open for the same defect and is the stronger of the two, so I would rather not merge this one.

#1252 also looks the binary up on the `PATH` before spawning, which matters because a shell reports an unknown command as a non-zero exit rather than as a spawn `error` event, so on Windows the handler added here never fires and a missing Maven still looks like a failed build. It carries a user-facing diagnostic and four tests as well; this PR has no test.

Assigning to @yegor256 to pick one and close the other.

---

